PR #12, opened 2020-05-02
add amsmath extension

chrisjsewell:
Adds extension to specifically capture top-level [amsmath](https://ctan.org/pkg/amsmath?lang=en) latex environments

cc @choldgraf @phaustin for comment

### DIFF
--- a/markdown_it/extensions/amsmath/__init__.py
+++ b/markdown_it/extensions/amsmath/__init__.py
@@ -28,6 +28,9 @@ ENVIRONMENTS = [
     "Bmatrix",
     "vmatrix",
     "Vmatrix",
+    # eqnarray is another math environment, it is not part of amsmath,
+    # and note that it is better to use align or equation+split instead
+    "eqnarray",
 ]
 # other "non-top-level" environments:
 
@@ -39,9 +42,6 @@ ENVIRONMENTS = [
 # 3.7 variants gathered, aligned,and alignedat are provided
 # whose total width is the actual width of the contents;
 # thus they can be used as a component in a containing expression
-
-# eqnarray is another math environment, but is not part of amsmath,
-# and it is better to use align or equation+split instead
 
 RE_OPEN = re.compile(r"\\begin\{(" + "|".join(ENVIRONMENTS) + r")([\*]?)\}")
 
--- a/markdown_it/extensions/amsmath/__init__.py
+++ b/markdown_it/extensions/amsmath/__init__.py
@@ -1,0 +1,103 @@
+"""An extension to capture ams math environments."""
+import re
+
+from markdown_it import MarkdownIt
+from markdown_it.rules_block import StateBlock
+from markdown_it.common.utils import escapeHtml
+
+# Taken from amsmath version 2.1
+# http://anorien.csc.warwick.ac.uk/mirrors/CTAN/macros/latex/required/amsmath/amsldoc.pdf
+ENVIRONMENTS = [
+    # 3.2 single equation with an automatically gen-erated number
+    "equation",
+    # 3.3 variation equation, used for equations that don’t fit on a single line
+    "multline",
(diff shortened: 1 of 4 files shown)


phaustin:
@chrisjsewell do I need to do anything special to activate the extension from myst_nb?

chrisjsewell:
> do I need to do anything special to activate the extension from myst_nb?

Few more steps until you can use it in myst-nb I'm afraid: https://github.com/executablebooks/MyST-NB/issues/126#issuecomment-622935821

phaustin:
this set  looks great -- is there a reason that the vertical, bracketed and parenthetical matrix environments (bmatrix, pmatrix, vmatrix) are missing?  Those would be at the top of the next tier of my own list of most commonly used mathjax environments from  the bottom section   of https://docs.mathjax.org/en/v2.7-latest/tex.html

markdown_it/extensions/amsmath/__init__.py
```diff
@@ -0,0 +1,103 @@
...
+# other "non-top-level" environments:
+
+# 3.4 the split environment is for single equations that are too long to fit on one line
+# and hence must be split into multiple lines,
+# it is intended for use only inside some other displayed equation structure,
+# usually an equation, align, or gather environment
+
+# 3.7 variants gathered, aligned,and alignedat are provided
+# whose total width is the actual width of the contents;
+# thus they can be used as a component in a containing expression
```
chrisjsewell, line 33: In response to your question: http://anorien.csc.warwick.ac.uk/mirrors/CTAN/macros/latex/required/amsmath/amsldoc.pdf states that these should not be top-level environments. Have you found a use case where they are?

chrisjsewell:
> bmatrix, pmatrix, vmatrix

Ah yes, I missed these in section 4.1 of http://anorien.csc.warwick.ac.uk/mirrors/CTAN/macros/latex/required/amsmath/amsldoc.pdf. That's why I asked for your comment 😄

phaustin:
just to flag the comment:

```
# eqnarray is another math environment, but is not part of amsmath,
# and it is better to use align or equation+split instead
```
Note that eqnarray is the default way to write systems of equations in latex.  It's definitely not recommended -- i.e.  http://tug.org/TUGboat/tb33-1/tb103madsen.pdf, but unfortunately it's going to be present in a lot of jupyter notebooks, and in all standard latex manuals.  AMS evidently decided that simply deprecating it in mathjax wasn't worth the struggle -- I think forcing users to take their medicine isn't a bad thing, but I'd anticipate a lot of bug reports from new users, which will need to be redirected to a standard doc snippet.

phaustin:
One other thing, I see flalign on your list, but it doesn't look like it's an environment either mathjax or katex support.  That would be something that would work for the latex builder, but revert to regular align for html?

chrisjsewell:
> Note that eqnarray is the default way to write systems of equations in latex

Added in 9d22992, for users who want to be stupid 😉

chrisjsewell:
> I see flalign on your list, but it doesn't look like it's an environment either mathjax or katex support.

Yeh strange that they don't. I think it should be kept here, since the goal is to capture all top-level amsmath environments, then handle it in some way as you mention upstream in myst-parser.
There are open issue for it on both repos:
- https://github.com/KaTeX/KaTeX/issues/1990
- https://github.com/mathjax/MathJax/issues/1075

Note a another thing to consider, which again mathjax and katex don't currently support but are good for latex PDF, would be the environments in http://mirror.ox.ac.uk/sites/ctan.org/macros/latex/contrib/breqn/breqn.pdf. Probably add this as an enhancement issue somewhere in executablebooks

phaustin:
Makes sense. How much effort will it be to handle something like https://mhchem.github.io/MathJax-mhchem ?

choldgraf:
way to go!

PR #22, opened 2020-07-20
Move the port.yaml files into correct folders

rowanc1:
Moved the ports for footnotes/front-matter into the correct folders.


chrisjsewell:
good catch 😄

Issue #8, opened 2020-03-31
Fix the more obscure backslash escape tests


chrisjsewell:
These are the final tests that need to be fixed; one that is directly applicable to CommonMark compliance, and the others are for additional compliance with markdown-it. They are all to do with how `\` escapes are treated for some of the more obscure cases in link titles / code fence languages, e.g. `[](<\&quot;> "\&amp;\&ouml;")`

- [x] https://github.com/ExecutableBookProject/markdown-it-py/blob/bc799d47e219fac146dd7f38f30641fb295a16fd/tests/test_cmark_spec/test_spec.py#L23
- [ ] https://github.com/ExecutableBookProject/markdown-it-py/blob/bc799d47e219fac146dd7f38f30641fb295a16fd/tests/test_port/test_fixtures.py#L28
- [ ] https://github.com/ExecutableBookProject/markdown-it-py/blob/bc799d47e219fac146dd7f38f30641fb295a16fd/tests/test_port/test_fixtures.py#L53
- [ ] https://github.com/executablebooks/markdown-it-py/pull/140/commits/f0c96ae5dbbbf0d27ed44e2f6ac237662a57001d (from https://github.com/markdown-it/markdown-it/commit/f156ed1bffec0e02b2353abf248c5b10c6101524)

chrisjsewell:
Note the portability issue lies in the fact that markdown-it uses the external JS dependencies [mdurl](https://www.npmjs.com/package/mdurl) and [punycode](https://www.npmjs.com/package/punycode) for their functions, which are obviously not directly useable for python.

chrisjsewell:
For the commonmark spec:

Backslash-escapes do not work inside autolinks:

```````````````````````````````` example
<http://example.com/\[\>
.
<p><a href="http://example.com/%5C%5B%5C">http://example.com/\[\</a></p>
````````````````````````````````

However, we get:

```
<p><a href="http://example.com/%5B%5C">http://example.com/\[\</a></p>
```

chrisjsewell:
@hukkinj1 and @tsutsu3, given your recent work, perhaps you have an idea of how to solve these?
It doesn't seem that anyone has really come across issues with these, but it would be nice to solve if possible before a v1.0.0 release

hukkinj1:
:+1: not picking this up right now, but I'll leave a comment here if I do later. Do these tests exist in JS markdown-it? Without the `skip`s? If yes, then the JS tests could be useful for debugging.

[...]

PR #36, opened 2020-08-17
🐛 FIX: empty lines after certain lists raises exception

sildar:
This is a minimal fix for #31 

Here's a copy/paste of things discussed in the issue for reference:

When processing lists, there is a call to tokenize() that advances the state.line attribute :

    list.py 
    line 264: state.md.block.tokenize(state, startLine, endLine)

After this line is executed, state.line can be larger than endLine, leading to the IndexError when checking the state at that index.

Breaking right after this line if state.line > endLine doesn't work though, we have to update the `nextline` variable before, as well as closing the list. It's actually already in the codebase, but one line too late:

    line 280 onwards:
        token = state.push("list_item_close", "li", -1)
        token.markup = chr(markerCharCode)

        nextLine = startLine = state.line
        itemLines[1] = nextLine
        contentStart = state.bMarks[startLine]

        if nextLine >= endLine:
            break

The easy solution would be to just put the check a few lines earlier.

    line 280 onwards:
        token = state.push("list_item_close", "li", -1)
        token.markup = chr(markerCharCode)

        nextLine = startLine = state.line  # we actually need to update these before breaking

        if nextLine >= endLine:
            break

        itemLines[1] = nextLine
        contentStart = state.bMarks[startLine]

Also, I'm not sure about what itemLines does, it's not really used at all in this method.

I looked at the tokenize() method mentioned at the beginning of this post and I don't think we should modify it.

### DIFF
--- a/markdown_it/rules_block/list.py
+++ b/markdown_it/rules_block/list.py
@@ -282,10 +282,11 @@ def list_block(state: StateBlock, startLine: int, endLine: int, silent: bool):
 
         nextLine = startLine = state.line
         itemLines[1] = nextLine
-        contentStart = state.bMarks[startLine]
 
         if nextLine >= endLine:
             break
+
+        contentStart = state.bMarks[startLine]
 
         #
         # Try to check if list is terminated or continued.


chrisjsewell:
Thanks @sildar 

Firstly, I think it would be good to add an (initially failing) test here to capture the issue.

Secondly, we should bear in mind the code in:
https://github.com/markdown-it/markdown-it/blob/master/lib/rules_block/list.js

`contentStart = state.bMarks[startLine]` is in the correct place, BUT, JS just returns `undefined` instead of raising an `IndexError`, for indices larger than the array. So yeh I think this solution should be fine, or a try/except IndexError block.

> Also, I'm not sure about what itemLines does, it's not really used at all in this method.

Its set here, as an alias for token.map: https://github.com/executablebooks/markdown-it-py/blob/8f168055d289d30e46202d0deb94ee0835e9f45f/markdown_it/rules_block/list.py#L232

I don't see any obvious reason not to just use `token.map`, but that it is what's done in the JS code 😬

chrisjsewell:
I added the test, so I think this is good to go.
Feel free to keep any other PRs coming though 😄 (#8 would be amazing, but it is real head-scratcher!)

PR #45, opened 2020-09-03
🐛 Fix: undefined variable in reference block

asmeurer:
I have a large document that reproduces this error, but I'm not sure how to
reproduce it in a shorter way. Perhaps someone more familiar with the code can
come up with a test case.

### DIFF
--- a/markdown_it/rules_block/reference.py
+++ b/markdown_it/rules_block/reference.py
@@ -74,6 +74,7 @@ def reference(state: StateBlock, startLine, _endLine, silent):
     string = state.getLines(startLine, nextLine, state.blkIndent, False).strip()
     maximum = len(string)
 
+    labelEnd = -1
     pos = 1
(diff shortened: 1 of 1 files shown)


chrisjsewell:
thanks @asmeurer, yeh looking at https://github.com/markdown-it/markdown-it/blob/master/lib/rules_block/reference.js, `labelEnd` is declared (but undefined) at the start of the function.
But then it "utilises" it doesn't fail because of the wrong IMO behaviour that `console.log(undefined<0)` resolves to `false` rather than throwing an incompatible type error.

chrisjsewell:
I made a slight change to your PR, I assume this will still fix the bug?

As its clearly a porting issue, I'll forgo asking for a test case, if you can't find an easy one

asmeurer:
I can confirm this still fixes the issue.

Issue #50, opened 2020-09-21
Issue with nested block-quote md syntax


mmcky:
**Describe the bug**

The following markdown snippet causes `jupytext` conversion to fail with error (`IndexError: list index out of range`)

```
> `!conda install quantecon`
```

Is this an issue with nested literals?

It appears to be valid `md` using https://spec.commonmark.org/dingus/. 

**To Reproduce**

Add snippet to any `md` document

run

`jupytext doc.md --to ipynb`

**Expected behavior**

successful transfer to `ipynb`

**Environment**

- `markdown_it_py = 0.5.4`

choldgraf:
does markdown-it-py parse that string properly as well? Does this error *only* pop up when used in conjunction with jupytext?

chrisjsewell:
I'll copy the traceback you added on slack:

```python
  File "/Users/matthewmckay/anaconda3/envs/tomyst/lib/python3.8/site-packages/markdown_it/parser_block.py", line 58, in tokenize
    state.line = line = state.skipEmptyLines(line)
  File "/Users/matthewmckay/anaconda3/envs/tomyst/lib/python3.8/site-packages/markdown_it/rules_block/state_block.py", line 134, in skipEmptyLines
    if (self.bMarks[from_pos] + self.tShift[from_pos]) < self.eMarks[from_pos]:
IndexError: list index out of range
```

This is likely due to a porting issue in JS->Python whereby markdown-it works IMO a bit through luck rather than judgement in some places, in that JS does not raise an `IndexError` when retrieving an index larger than the length of the array but instead just returns `undefined`, and on top of that does not raise a `TypeError` when evaluating undefined against a number, e.g.

```javascript
>> a = [1, 2, 3]
>> console.log(a[4])
undefined
>> console.log(a[4] > 1)
false
```

so the fix will be something along the lines of:

```python
if from_pos < len(self.eMarks) and (self.bMarks[from_pos] + self.tShift[from_pos]) < self.eMarks[from_pos]:
```

chrisjsewell:
@mmcky I've created a PR just with the snippet you gave, but this passes: #51

Can you work out the minimal preceding/following text from your full file, which can recreate the `IndexError`?

(tip: if you recreate this PR, you can just run `tox -- -k #50` to run the single test)

mmcky:
[...]

PR #51, opened 2020-09-21
🐛 FIX: #50

chrisjsewell:


mmcky:
thanks @chrisjsewell this works on my machine also (even with two trailing newlines) -- so I guess this can just be merged as a useful test case.

chrisjsewell:
fixed in #61

Issue #53, opened 2020-10-01
ENH: CLI interface to parse markdown and render HTML


westurner:
**Is your feature request related to a problem? Please describe.**

Use case: Render a snippet of markdown in a file on disk into HTML.

**Describe the solution you'd like**

A commandline interface to parse markdown and render HTML (and optionally configure markdown-it-py first)

**Describe alternatives you've considered**

I could write a script with an `argparse.ArgumentParser` and use that; but that might as well just be a `console_scripts` `entry_point` in markdown-it-py; in which case an actual `argparse.ArgumentParser` is justifiied.

**Additional context**

None

chrisjsewell:
Heya, note there is already a CLI interface (accepts a file, stdin, or interactive mode):

```console
$ markdown-it
markdown-it-py [version 0.1.0] (interactive)
Type Ctrl-D to complete input, or Ctrl-C to exit.
>>> > **hallo** there!
...
<blockquote>
<p><strong>hallo</strong> there!</p>
</blockquote>
```

Its not particularly well documented though, and it can't be configured, so indeed these could be improved.

westurner:
I should've looked more closely for the entry_point in setup.py.

- [ ] DOC: cli.parse: Add description and maybe also epilog with usage examples
  https://docs.python.org/3/library/argparse.html#description
- [ ] DOC: README.md: Include output of `markdown-it -h` under e.g. Usage > CLI

westurner:
- [ ] ENH: Add output path(s?
- [ ] ENH: Add way to do configuration
  - Discussion: INI / TOML / YAML / Python and/or lots of `--args-to-test`

trend-jannis-weigend:
On Windows `Ctrl-Z` is required instead of `Ctrl-D` to complete input.

Dobatymo:
Windows requires `Ctrl-Z` instead of `Ctrl-D` to complete input.

PR #55, opened 2020-10-01
🐛 FIX: cli.parse: return 0 for non-error #53

westurner:
This is necessary so that make doesn't fail to execute this task due to `markdown-it` returning True (1) instead of zero (0) for success:

```makefile
header-footer-templates:
	@#pip install markdown-it-py
	markdown-it README.md.header.md > index.html.header.html
	markdown-it README.md.footer.md > index.html.footer.html

```


tests/test_cli.py
```diff
@@ -2,14 +2,21 @@
...
     with tempfile.TemporaryDirectory() as tempdir:
         path = pathlib.Path(tempdir).joinpath("test.md")
         path.write_text("a b c")
-        assert parse.main([str(path)])
+        assert parse.main([str(path)]) == 0
+
+
+def test_parse_fail():
+    with pytest.raises(SystemExit):
+        assert parse.main(["/tmp/nonexistant_path/for_cli_test.md"]) == 1
```
chrisjsewell, line 19: nitpick, but surely you will never actually test this assert, because the parse raises a `SystemExit`

westurner: Yeah, but it's helpful to document that the returncode when SystemExit is raised is 1.

Is there branch coverage on the tests/?

westurner: I've removed the assert.

https://docs.python.org/3/library/sys.html#sys.exit
https://docs.python.org/3/library/exceptions.html#SystemExit

tests/test_cli.py
```diff
@@ -2,14 +2,21 @@
...
     with tempfile.TemporaryDirectory() as tempdir:
         path = pathlib.Path(tempdir).joinpath("test.md")
         path.write_text("a b c")
-        assert parse.main([str(path)])
+        assert parse.main([str(path)]) == 0
+
+
+def test_parse_fail():
+    with pytest.raises(SystemExit):
+        parse.main(["/tmp/nonexistant_path/for_cli_test.md"])
```
hukkinj1, line 19: SystemExit per se doesn't mean that exit code is non-zero. We can assert that with something like:
```suggestion
    with pytest.raises(SystemExit) as exc_info:
        parse.main(["/tmp/nonexistant_path/for_cli_test.md"])
    assert exc_info.value.code == 1
```

tests/test_cli.py
```diff
@@ -2,14 +2,22 @@
...
         path = pathlib.Path(tempdir).joinpath("test.md")
         path.write_text("a b c")
-        assert parse.main([str(path)])
+        assert parse.main([str(path)]) == 0
+
+
+def test_parse_fail():
+    with pytest.raises(SystemExit) as exc_info:
+        parse.main(["/tmp/nonexistant_path/for_cli_test.md"])
+    assert exc_info.value.code == 1
```
hukkinj1, line 20: One way to fix the test failure is:
```suggestion
    assert isinstance(exc_info.value.code, str)
```
---
Another way, that I personally would prefer but don't know if @chrisjsewell would, is go to `markdown_it.cli.parse.convert_file` and replace
```python
sys.exit('Cannot open file "{}".'.format(filename))
```
with
```python
sys.stderr.write('Cannot open file "{}".\n'.format(filename))
sys.exit(1)
```

chrisjsewell:
Superseded by #124

westurner:
Thanks

PR #57, opened 2020-10-10
🧪 Make `read_fixture_file` always read UTF-8

hukkinj1:
Currently `read_fixture_file` decides file encoding based on platform. It may read different text from the same file when running on a different platform (Windows, linux, macOS etc.), and as a result, break tests on some platforms (typically windows, lol).

This PR makes the function read UTF-8 regardless of platform.


chrisjsewell:
cheers!

PR #58, opened 2020-10-16
👌 IMPROVE: Use `re.sub()` instead of `re.subn()[0]`

hukkinj1:


chrisjsewell:
looks good to me thanks @hukkinj1

Issue #19, opened 2020-05-18
MyPy Typechecking


0az:
As part of getting to learning how to hack on the package, I've started running MyPy for a "guided tour" of the API.

Do you mind if I send in one or more PRs to fix/silence some of the warnings?

chrisjsewell:
Yeh go for it 👍

0az:
Something that I noticed when walking through the code:

There's a lot of untyped `dict`s and `AttrDicts` that probably can get replaced with dataclasses/attrs/namedtuple/etc.

Example:

```
def postProcess(state: StateInline):
    """Walk through delimiter list and replace text tokens with tags."""
    tokens_meta = state.tokens_meta
    maximum = len(state.tokens_meta)
    _postProcess(state, state.delimiters)

    curr = 0
    while curr < maximum:
        try:
            tokens_meta[curr]
        except IndexError:
            pass
        else:
            if tokens_meta[curr] and "delimiters" in tokens_meta[curr]:
                _postProcess(state, tokens_meta[curr]["delimiters"])
        curr += 1
```
`tokens_meta` can probably get swapped over to a TypedDict, or something like:
```
@dataclass
class TokenMetadata:
	delimiters: List[Delimiter]
```

The downside of TypedDict is that it's a 3.8+ feature, and some functionality requires `typing_extensions` until 3.9. One workaround could be using `__getitem__`/etc.

This is out of scope for what I intend, though.

0az:
@chrisjsewell Which test failures should I expect?

chrisjsewell:
Which test failures should I expect?

None? 
I'm not sure if I understand what you, all the tests currently: https://github.com/executablebooks/markdown-it-py/actions/runs/105415163
Why would you expect some to fail?

FYI in-terma of mypy, you could just try add the pre-commit hook to the pre-commit config file, then fix the code accordingly: https://github.com/pre-commit/mirrors-mypy

0az:
Ah, sorry. I got a bug I introduced locally mixed up with some of the skipped tests.

Right now, there's around ~70 standing errors that I still need to fix, after grabbing the lowest hanging fruit, so I'll need to figure something out before adding a pre-commit hook.

chrisjsewell:
Yeh no worries. The only thing to bear in mind is that we "shouldn't" (without very good reason) change the code in any way that deviates from the *source* [markdown-it](https://github.com/markdown-it/markdown-it) API.

[...]

PR #64, opened 2020-10-21
🔧 MAINTAIN: Add mypy type-checking

hukkinj1:
Closes https://github.com/executablebooks/markdown-it-py/issues/19

### DIFF
--- a/markdown_it/common/utils.py
+++ b/markdown_it/common/utils.py
@@ -53,7 +53,7 @@ def assign(obj):
     #   })
     # })
 
-    return obj
+    # return obj
 
 
 def arrayReplaceAt(src: list, pos: int, newElements: list):
@@ -139,9 +139,10 @@ def replaceEntityPattern(match, name):
 
 
 def unescapeMd(string: str):
-    if "\\" in string:
-        return string
-    return string.replace(UNESCAPE_MD_RE, "$1")
+    raise NotImplementedError
+    # if "\\" in string:
+    #     return string
+    # return string.replace(UNESCAPE_MD_RE, "$1")
 
 
 def unescapeAll(string: str):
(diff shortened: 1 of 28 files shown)


markdown_it/common/utils.py
```diff
@@ -141,9 +141,10 @@ def replaceEntityPattern(match, name):
 
 
 def unescapeMd(string: str):
-    if "\\" in string:
-        return string
-    return string.replace(UNESCAPE_MD_RE, "$1")
+    raise NotImplementedError
+    # if "\\" in string:
+    #     return string
+    # return string.replace(UNESCAPE_MD_RE, "$1")
```
hukkinj1, line 147: I simply commented this function out, because `str.replace()` takes strings as input, not re.Patterns, so this is probably broken.

chrisjsewell:
I will get round to looking at this eventually! Just want to make sure I look thoroughly, since its touching core code

hukkinj1:
I rebased on master, and added one more commit that adds `no_implicit_optional = True` mypy configuration. This is to conform to [updated PEP 484](https://github.com/python/peps/pull/689). Eventually this will be default mypy configuration too.

chrisjsewell:
👍 

> this will be default mypy configuration too.

TBH it would be so much nicer and concise if you could use the typescript `?` notation:

```python
def func(a: Employee? = None):
  pass
```

hukkinj1:
I think I've seen some discussion around that `Optional` syntax somewhere. Python typing is moving forward at quite a rapid pace so maybe one day we'll see that. In the next Python release we already get [union types with `TypeX | TypeY`](https://www.python.org/dev/peps/pep-0604/) which I really like :)

chrisjsewell:
Oh thats nice 😄 Yeh for typing, I think they should just copy everything from typescript, which they are basically gradually doing

hukkinj1:
Haha yeh seems so. Btw. the new union syntax obviously lets us do `int | None` which is more concise than `Optional[int]` and requires no import from `typing`, so partially solves the problem.

chrisjsewell (approved):

PR #65, opened 2020-10-22
🧪 TESTS: Add PyPy3 to matrix

hukkinj1:


choldgraf:
I am not sure what are the implications of this. Could you explain a bit the reasoning for adding this extra job to the test suite? What's the added benefit over testing the different packages of python? Do we expect these tests to be any more brittle or require more maintenance than others?

chrisjsewell:
Oh I was just gonna merge this, if the tests pass and it doesn't take too long to run, I see no problem to adding it

choldgraf:
sounds good 👍

hukkinj1:
> Could you explain a bit the reasoning for adding this extra job to the test suite?

The reasoning is, since there is a PyPy Trove classifier in the package's metadata, it's probably good practice back that up by testing it.

> Do we expect these tests to be any more brittle

PyPy builds typically break if the package or a dependency are not pure Python (which is not the currently not the case for this package).

chrisjsewell (approved):
cheers!

PR #69, opened 2020-11-03
🐛 FIX: `_Entities.__contains__`

hukkinj1:


chrisjsewell (approved):

hukkinj1:
@chrisjsewell btw, what do you think about replacing `markdown_it.common.entities.DATA` with
```python
import html
DATA = html.entities.html5
```

chrisjsewell:
> what do you think about replacing markdown_it.common.entities.DATA with

they only have 106 keys in common, why the big difference?

```python
len(set(DATA).intersection(set(html.entities.html5))) == 106
```

hukkinj1:
Ah, I see. `html.entities.html5` includes trailing semicolons whereas `DATA` does not. Explained here: https://docs.python.org/3/library/html.entities.html

Besides that I think the data should be the same.

chrisjsewell:
yep thats more like it 👍 :

```python
len(set([_+';' for _ in DATA]).intersection(set(html.entities.html5))) == len(DATA)
```

so yeh, if we account for this trailing `;` in some fashion

PR #73, opened 2020-11-04
🔧 MAINTAIN: Don't allow `None` value for `StateBase.src`

hukkinj1:
Javascript implementation doesn't handle null values for `StateBase.src`, so I think we shouldn't either.

Also, it seems the `None` handling isn't fully implemented in all `StateBase` subclasses. E.g. `StateInline.__init__` will throw an exception when src is None (from the line `self.posMax = len(self.src)`) even though there is None handling in the `__init__` function before that.


chrisjsewell:
when you say the javascript doesn't handle it, you do know that javascript handles indexing very differently to python?
As I explained here https://github.com/executablebooks/markdown-it-py/issues/50#issuecomment-696240266, does not raise IndexErrors when indexing None or and index that is out of bounds.

You should also look at #32, where this ordinal caching was introduced, and where we specifically introduced the catch of `None` to fix a bug: https://github.com/executablebooks/markdown-it-py/pull/32#issuecomment-674295455

hukkinj1:
> you do know that javascript handles indexing very differently to python?

Yeah, I'm no JS expert :smile:, but hopefully do know some basics.

> does not raise IndexErrors when indexing None or and index that is out of bounds.

Out of bounds will not raise for sure, but I'm pretty sure indexing `null` does, no?

For instance this line https://github.com/markdown-it/markdown-it/blob/1b204ef54063aa7fde055f0e801e234be0a74d1e/lib/rules_inline/state_inline.js#L20 in the JS implementation will surely raise if `src` is `null` (accessing `length` atribute of `null`).

What would be the use case for initializing a State with `src = None`, and if there is one, could that perhaps be replaced by `src = ""`?

If `None` support needs to be kept then I think the type annotations should probably be changed to reflect that, and at least [this line](https://github.com/executablebooks/markdown-it-py/blob/b533d6d787462057061440e00a5da063798347f5/markdown_it/rules_inline/state_inline.py#L55) needs to be able to handle a `None`.

hukkinj1:
I found the MyST-NB usage of `StateCore(None, ...`. All tests seem to pass also when instantiating with an empty string: https://github.com/executablebooks/MyST-NB/pull/277 so it seems at least that particular use doesn't have a hard requirement for `src = None`.

hukkinj1:
I think this should be a pretty safe merge now that myst-nb has seen a few releases after the usage of `StateBase.src = None` was removed there, and jupyter book release also uses a myst-nb version that doesn't set the `None`. What do you think @chrisjsewell ?

Issue #75, opened 2020-11-10
Plugin for task lists


wna-se:
**Is your feature request related to a problem? Please describe.**

Ultimately I would like to add support for [GitHub style rendering of task lists](https://docs.github.com/en/free-pro-team@latest/github/managing-your-work-on-github/about-task-lists) using \[ \] and \[x\] in a downstream repository of executablebooks – e.g. jupyter-book – and I've tracked down this repository as a candidate for a pull request. 

**Describe the solution you'd like**

Porting a plugin for markdown-it such as [@hackmd/markdown-it-task-lists](https://www.npmjs.com/package/@hackmd/markdown-it-task-lists), source at  [revin/markdown-it-task-lists](https://github.com/revin/markdown-it-task-lists) to the extensions folder in this repository.

**Describe alternatives you've considered**

Creating a separate repository for the plugin and adding configuration options for external plugins to the downstream repositories MyST-Parser, MyST-NB and jupyter-book. This is a bigger task than I can do by myself.

**Additional context**

Se pull request.

chrisjsewell:
thanks I will try to get round to looking to add this to myst-parser (and thus jupyter-book) soon (probably once #78 is merged)

PR #76, opened 2020-11-10
✨ NEW: Add markdown-it-task-lists plugin

wna-se:
Illustrates a solution to #75

@hackmd/markdown-it-task-lists:
https://github.com/revin/markdown-it-task-lists


chrisjsewell:
Awesome thanks. On an initial browse it's looking good 😀
Can you add a port.yaml, as with the other extensions, detailing the current version, commit etc of the ported package, so that it's easy to check against at a later date

wna-se:
@chrisjsewell Thanks for reviewing. I've added a port.yaml file.

chrisjsewell:
@wna-se you know you can just use pre-commit locally for the formatting and lint checks?

```
pip install pre-commit
pre-commit run --all
```

chrisjsewell:
oh and also can you add the plugin to https://github.com/executablebooks/markdown-it-py/blob/master/docs/plugins.md (the same as how its done for the other plugins)

markdown_it/extensions/tasklists/__init__.py
```diff
@@ -0,0 +1,146 @@
...
+
+from markdown_it.token import Token
+
+
+def tasklists_plugin(
+    md: MarkdownIt,
+    enabled: bool = False,
+    label: bool = False,
+    label_after: bool = False,
+):
```
chrisjsewell, line 33: nearly there!

If you could add a docstring here, explaining the plugin and the parameters, so that it will show up in the documentation.

something like

```suggestion
def tasklists_plugin(
    md: MarkdownIt,
    enabled: bool: bool = False,
    label: bool = False,
    label_after: bool = False,
):
   """A plugin which parses GitHub style checkbox TODO lists.
   
    For example::
   
       - [ ] An item that needs doing
       - [x] An item that is complete
   
   :param enabled: this does...
   :param label: this does...
   :param label_after: this does...

   """
```

chrisjsewell (changes requested):
see comment

chrisjsewell:
I added a commit to your branch, which should hopefully fix the docs fail

markdown_it/extensions/tasklists/__init__.py
chrisjsewell: oh you did this just as I was typing this lol.
Anyhow, perhaps adding the actual example above would be good

wna-se:
Thanks! Though I can't find the docstring in the rendered version of the docs for this pull request. Did something go wrong?

chrisjsewell:
> Thanks! Though I can't find the docstring in the rendered version of the docs for this pull request. Did something go wrong?

Yeps its all there 👍 : on this page: https://markdown-it-py--76.org.readthedocs.build/en/76/plugins.html

wna-se:
> > Thanks! Though I can't find the docstring in the rendered version of the docs for this pull request. Did something go wrong?
> 
> Yeps its all there 👍 : on this page: https://markdown-it-py--76.org.readthedocs.build/en/76/plugins.html

When I access that link I can find the “heading” for the tasklists plugin but right below I suspect that there should also be a description of its function and parameters, which I added to the docstring.

markdown_it/extensions/tasklists/__init__.py
```diff
@@ -0,0 +1,159 @@
...
+    def todoify(token: Token, token_constructor):
+        token.children.insert(0, make_checkbox(token, token_constructor))
+        token.children[1].content = token.children[1].content[3:]
+        token.content = token.content[3:]
+
+        if use_label_wrapper:
+            if use_label_after:
+                token.children.pop()
+
+                # Use large random number as id property of the checkbox.
```
chrisjsewell, line 97: just a nitpick, but really it would be better to use https://docs.python.org/3/library/uuid.html here, e.g.:

```python
from uuid import uuid4

checklist_id = f"task-item-{uuid4()}"
```

markdown_it/extensions/tasklists/__init__.py
```diff
@@ -0,0 +1,159 @@
...
+    label: bool = False,
+    label_after: bool = False,
+):
+    """Plugin for building task/todo lists out of markdown lists with items starting with [ ] or [x]
+    .. Nothing else
+
+    The rendered checkboxes are disabled; to change this, pass a truthy value into the enabled
+    property of the plugin options.
+
+    :param enabled: True enables the rendered checkboxes
```
chrisjsewell, line 40: it wasn't really clear to me what enable meant in the current description

```suggestion
    :param enabled: True adds an "enabled" CSS class to each list item 
```

chrisjsewell: oh no actually maybe this was a misunderstanding by me.

I see it adds `disabled="disabled"` to the `input` markup, which I guess means it is frozen and you can't toggle its value

chrisjsewell:
> When I access that link I can find the “heading” for the tasklists plugin but right below I suspect that there should also be a description of its function and parameters, which I added to the docstring.

I see:

image

markdown_it/extensions/tasklists/__init__.py
wna-se: I agree, and I had the same thought but chose to port the plugin ”as is”.

markdown_it/extensions/tasklists/__init__.py
wna-se: Yepp, it's the ability to change the values of the checkboxes. I ported the documentation with the phasing mostly intact.

wna-se:
> > When I access that link I can find the “heading” for the tasklists plugin but right below I suspect that there should also be a description of its function and parameters, which I added to the docstring.
> 
> I see:
> 
> image

Must've been a browser cache issue, sorry about the confusion.

wna-se:
Should I fix these issues here – metaphorically ”forking” the original plugin or add the suggestion to the upstream plugin first?

markdown_it/extensions/tasklists/__init__.py
wna-se: I added the change and a comment in port.yml

markdown_it/extensions/tasklists/__init__.py
wna-se: There is a comment on this option in the general part of the docstring

markdown_it/extensions/tasklists/__init__.py
wna-se: Good idea, I added the example to the docstring.

chrisjsewell (approved):

PR #77, opened 2020-11-12
🐛 FIX: `MarkdownIt.render`; don't copy empty `env` arg

hukkinj1:
This PR makes `MarkdownIt.render` not copy an `env` if it is an empty `AttrDict`, making it possible for a caller to capture changes made to `env`. This is in line with how `MarkdownIt.parse` method, and the JS implementation works.


chrisjsewell:
cheers

PR #78, opened 2020-11-15
✨ NEW: Add linkify

### LINKED ISSUE #5: Finalise port of non-standard rules to python
These are just non-standard additional rules:

- [x] replacements (Simple typographic replacements)
- [x] smartquotes (Convert straight quotation marks to typographic ones)
- [ ] linkify (Replace link-like texts with link nodes.)

They are currently in the repo as their original `.js` files

tsutsu3:
close #5

Implement `linkify` with [linkify-it-py](https://github.com/tsutsu3/linkify-it-py)

### Implementation memo

- ~~If `likify-it-py` is not installed, import `dummy LinkifyIt`~~
- Throw an exception only when `linkify` rule enabled, `linkify` option is `True` and  `likify-it-py` is not installed
- The exceptions raise at parse time, not at initialization time
- Avoided including `likify-it-py` in dependency package


chrisjsewell:
The final piece of the puzzle!
Thanks 🙏 I should be able to have a look later today

tsutsu3:
Is this implementation good?

If it's OK, increase coverage

chrisjsewell:
Well it passes all the markdown-it fixture tests. So that's always a good start 😀

@hukkinj1 might also want to have a look?

markdown_it/extra/__init__.py
```diff
@@ -0,0 +1 @@
+from .linkify_it import LinkifyIt  # noqa: F401
```
hukkinj1, line 1:
```suggestion
```

markdown_it/extra/linkify_it.py
```diff
@@ -0,0 +1,43 @@
...
+    #     raise NotImportedError
+
+    # def test(self, text):
+    #     raise NotImportedError
+
+    # def test_schema_at(self, text, name, position):
+    #     raise NotImportedError
+
+    # def tlds(self, list_tlds, keep_old=False):
+    #     raise NotImportedError
```
hukkinj1, line 43:
```suggestion
```

markdown_it/main.py
```diff
@@ -11,6 +11,11 @@
 from .renderer import RendererHTML
 from .utils import AttrDict
 
+try:
+    from linkify_it import LinkifyIt
+except (ImportError, ModuleNotFoundError):
+    from markdown_it.extra import LinkifyIt
```
hukkinj1, line 17:
```suggestion
try:
    from linkify_it import LinkifyIt
except ImportError:
    LinkifyIt = None
```

markdown_it/main.py
```diff
@@ -41,8 +46,7 @@ def __init__(
         self.options = {}
         self.configure(config)
 
-        # var LinkifyIt    = require('linkify-it')
-        # self.linkify = LinkifyIt()  # TODO maybe see https://github.com/Suor/autolink
+        self.linkify = LinkifyIt()
```
hukkinj1, line 49:
```suggestion
        self.linkify = LinkifyIt() if LinkifyIt else None
```

markdown_it/rules_core/linkify.py
```diff
@@ -0,0 +1,127 @@
...
+def isLinkClose(string: str):
+    return LINK_CLOSE_RE.search(string)
+
+
+def linkify(state: StateCore):
+    blockTokens = state.tokens
+
+    if not state.md.options.linkify:
+        return
+
```
hukkinj1, line 30:
```suggestion
    if not state.md.linkify:
        raise Exception("Linkify enabed but not installed")
```

hukkinj1 (changes requested):
Nice! I would suggest a change in how the "linkify enabled but not installed" case is handled, that should result in overall a bit less code, and avoids giving `LinkifyIt.pretest` a secondary task.

tsutsu3:
> Nice! I would suggest a change in how the "linkify enabled but not installed" case is handled, that should result in overall a bit less code, and avoids giving `LinkifyIt.pretest` a secondary task.

Oh, thanks. :smile:

markdown_it/rules_core/linkify.py
```diff
@@ -0,0 +1,130 @@
...
+                i -= 1
+                while (
+                    tokens[i].level != currentToken.level
+                    and tokens[i].type != "link_open"
+                ):
+                    i -= 1
+                continue
+
+        # Skip content of html tag links
+        if currentToken.type == "html_inline":
```
tsutsu3, line 61: The depth of the Nest is wrong.

Should be the same as line 51.

markdown_it/rules_core/linkify.py
```diff
@@ -0,0 +1,130 @@
...
+            continue
+
+        tokens = blockTokens[j].children
+
+        htmlLinkLevel = 0
+
+        # We scan from the end, to keep position when new tags added.
+        # Use reversed logic in links start/end match
+        assert tokens is not None
+        for i in range(len(tokens))[::-1]:
```
hukkinj1, line 47: This is just a preference thing but I personally always found `reversed(iterable)` more readable than `iterable[::-1]`

markdown_it/rules_core/linkify.py
```diff
@@ -0,0 +1,130 @@
...
+
+
+def linkify(state: StateCore):
+    blockTokens = state.tokens
+
+    if not state.md.options.linkify:
+        return
+
+    if not state.md.linkify:
+        raise Exception("Linkify enabed but not installed.")
```
hukkinj1, line 32:
```suggestion
        raise Exception("Linkify enabled but not installed.")
```
There's a typo here which originates from my suggestion. Sorry! Btw, the text was just something I just quickly made up, dont know, maybe we want it to be more elaborate somehow @chrisjsewell ?

hukkinj1: Also, maybe it could be justified to make the exception type `ModuleNotFoundError` ?

markdown_it/rules_core/linkify.py
tsutsu3: This doesn't work correctly:
```
no blank line before text
github.com
```

markdown_it/rules_core/linkify.py
tsutsu3: or `ImportError`?

I'm worried about exception type. 🤔

hukkinj1: I think I'd vote for `ModuleNotFoundError` (a subclass of `ImportError`, a bit more precise) as it's equivalent to the root cause of the issue here (the exception we suppress in`markdown_it/main.py` line 16, and never need to raise unless this line runs).

markdown_it/rules_core/linkify.py
tsutsu3: `for` statement did not calculate the index `i` correctly.

So I changed to a `while` statement.

markdown_it/rules_core/linkify.py
tsutsu3: Not 'Outdated' but fixed.

markdown_it/rules_core/linkify.py
```diff
@@ -0,0 +1,143 @@
...
+        htmlLinkLevel = 0
+
+        # We scan from the end, to keep position when new tags added.
+        # Use reversed logic in links start/end match
+        assert tokens is not None
+        i = len(tokens)
+        while i >= 0:
+            i -= 1
+            if i < 0:
+                break
```
hukkinj1, line 51:
```suggestion
        while i >= 1:
            i -= 1
```
Wouldn't this do the same thing?

tsutsu3: Right, same 😲

chrisjsewell:
@hukkinj1 have you got any more notes on this?

hukkinj1 (dismissed):
This looks very good to me!

setup.py
```diff
@@ -67,6 +67,7 @@ def get_version():
             "mistletoe-ebp~=0.10.0",
             "panflute~=1.12",
         ],
+        "linkify": ["linkify-it-py"],
```
chrisjsewell, line 70:
```suggestion
        "linkify": ["linkify-it-py~=1.0"],
```

chrisjsewell (approved):
thanks @tsutsu3 😄

Issue #80, opened 2020-11-23
UnicodeError with some HTML entities 


lesteve:
**Describe the bug**

There is a rendering issue with some HTML entities for example this one `&#x1F4AC;` which is 💬 (speech balloon according to https://graphemica.com/%F0%9F%92%AC), creates a HTML string that can not be encoded in UTF8. This causes a hard to understand `UnicodeError` in JupyterBook.

For completeness the work-around of copying and pasting 💬 into my editor worked, unfortunately this was not the first thing I tried :sweat: ...

**To Reproduce**

```py
from markdown_it import MarkdownIt

md = MarkdownIt()

text = ("&#x1F4AC;")
html = md.render(text)
html.encode('utf8')
```

yields the error:
```
---------------------------------------------------------------------------
UnicodeEncodeError                        Traceback (most recent call last)
<ipython-input-8-063cca89ac1a> in <module>
      5 text = ("&#x1F4AC;")
      6 html = md.render(text)
----> 7 html.encode('utf8')

UnicodeEncodeError: 'utf-8' codec can't encode characters in position 3-4: surrogates not allowed
```

**Expected behavior**

The HTML entity is rendered correctly, for example the Javascript markdown-it parser renders `&#x1F4AC;` fine: see [this](
https://markdown-it.github.io/#md3=%7B%22source%22%3A%22%26%23x1F4AC%3B%5Cn%22%2C%22defaults%22%3A%7B%22html%22%3Afalse%2C%22xhtmlOut%22%3Afalse%2C%22breaks%22%3Afalse%2C%22langPrefix%22%3A%22language-%22%2C%22linkify%22%3Atrue%2C%22typographer%22%3Atrue%2C%22_highlight%22%3Atrue%2C%22_strict%22%3Afalse%2C%22_view%22%3A%22html%22%7D%7D)

**Additional context**

I was having the problem in JupyterBook when adding `&#x1F4AC;` into one of my `.md` file.

<details>

<summary>Full JupyterBook stack-trace</summary>

```
The full traceback has been saved in /tmp/sphinx-err-ssczzp1s.log, if you want to report the issue to the developers.
Traceback (most recent call last):
  File "/home/lesteve/miniconda3/envs/scikit-learn-tutorial/lib/python3.8/site-packages/jupyter_book/sphinx.py", line 141, in build_sphinx
    app.build(force_all, filenames)
[...]

PR #81, opened 2020-11-23
🐛 FIX: parsing of unicode ordinals

hukkinj1:
Closes https://github.com/executablebooks/markdown-it-py/issues/80

This seems like a porting issue to me: Javascript needs two string characters for codepoints larger than `0xFFFF` but Python 3 can represent any unicode codepoint in one character.

This PR removes the handling that splits a large codepoint to two chars.

### DIFF
--- a/markdown_it/common/utils.py
+++ b/markdown_it/common/utils.py
@@ -92,15 +92,7 @@ def isValidEntityCode(c):
     return True
 
 
-def fromCodePoint(c):
-
-    if c > 0xFFFF:
-        c -= 0x10000
-        surrogate1 = 0xD800 + (c >> 10)
-        surrogate2 = 0xDC00 + (c & 0x3FF)
-
-        return "".join(map(chr, [surrogate1, surrogate2]))
-
+def fromCodePoint(c: int) -> str:
     return chr(c)
 
 
--- a/markdown_it/common/utils.py
+++ b/markdown_it/common/utils.py
@@ -93,12 +93,12 @@ def isValidEntityCode(c):
 
 
 def fromCodePoint(c: int) -> str:
-   """Convert ordinal to unicode.
-   
-   Note, in the original Javascript two string characters were required,
-   for codepoints larger than `0xFFFF`.
-   But Python 3 can represent any unicode codepoint in one character.
-   """
+    """Convert ordinal to unicode.
+
+    Note, in the original Javascript two string characters were required,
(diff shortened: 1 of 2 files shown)


hukkinj1:
@chrisjsewell can you think of any other places in the codebase where this Javascript string (string of UTF-16 code units) versus Python3 string (string of Unicode codepoints) difference could cause issues.

I just realized we unintentionally fixed one instance of this in https://github.com/executablebooks/markdown-it-py/pull/70 :smile: The old entity data seemed to contain UTF-16 code units.

chrisjsewell:
Not that I know of, I'll have a quick look later

lesteve:
Just a gentle reminder about this one ;-). I completely understand if you have higher priorities of course.

chrisjsewell (changes requested):

markdown_it/common/utils.py
```diff
@@ -92,15 +92,7 @@ def isValidEntityCode(c):
...
-def fromCodePoint(c):
-
-    if c > 0xFFFF:
-        c -= 0x10000
-        surrogate1 = 0xD800 + (c >> 10)
-        surrogate2 = 0xDC00 + (c & 0x3FF)
-
-        return "".join(map(chr, [surrogate1, surrogate2]))
-
+def fromCodePoint(c: int) -> str:
```
chrisjsewell, line 95:
```suggestion
def fromCodePoint(c: int) -> str:
   """Convert ordinal to unicode.
   
   Note, in the original Javascript two string characters were required,
   for codepoints larger than `0xFFFF`.
   But Python 3 can represent any unicode codepoint in one character.
   """
```

markdown_it/common/utils.py
```diff
@@ -92,15 +92,13 @@ def isValidEntityCode(c):
...
+   """Convert ordinal to unicode.
+   
+   Note, in the original Javascript two string characters were required,
+   for codepoints larger than `0xFFFF`.
+   But Python 3 can represent any unicode codepoint in one character.
+   """
     return chr(c)
```
chrisjsewell, line 102:
```suggestion
    """Convert ordinal to unicode.

    Note, in the original Javascript two string characters were required,
    for codepoints larger than `0xFFFF`.
    But Python 3 can represent any unicode codepoint in one character.
    """
    return chr(c)
```

chrisjsewell (approved):

PR #93, opened 2020-12-14
🗑 DEPRECATE: Move plugins to `mdit_py_plugins`

chrisjsewell:
Plugins (in `markdown_it.extensions`) have now been moved to https://github.com/executablebooks/mdit-py-plugins.
This will allow for their maintenance to occur on a different cycle to the core code, facilitating the release of a v1.0.0 for this package


chrisjsewell:
@hukkinj1 I imagine this will affect you 😬

chrisjsewell:
oh and I just need to get mdit-py-plugins on conda before releasing

hukkinj1 (dismissed):
Thanks for the heads-up. I think I'll only need to patch `mdformat-gfm`.

Changes look good to me!

PR #95, opened 2020-12-15
🐛 FIX: `Token.attrs` type annotations

hukkinj1:
See https://github.com/executablebooks/markdown-it-py/blob/047a71f67ea8cf5c52faa1e42a3c9698a7274ab5/markdown_it/rules_block/list.py#L171

`Token.attrs` type is not strictly `List[List[str]]` (as the annotations currently imply). In the example line we assign an `int` value (`markerValue`). This PR annotates `Token.attrs` more loosely as `List[list]]`.


hukkinj1:
The most natural type here would probably be `List[Tuple[str, Any]]`, but that requires changing code, not only annotations. I can do that if it's ok to @chrisjsewell 

The upside of that change is that the typing would support `dict(token.attrs)` (mypy is happy with converting `Iterable[Tuple[Any, Any]]` to dict, but not `Iterable[list]`.

chrisjsewell:
> The most natural type for Token.attrs would probably be List[Tuple[str, Any]], but that requires changing code, not only annotations. I can do that if it's ok to @chrisjsewell

hmm, possibly, just have to think if there could be any side effects, e.g. to plugins

hukkinj1:
Made a PR to try out the `Tuple[str, Any]` change here https://github.com/executablebooks/markdown-it-py/pull/102 . Tests seem to pass. Next I'll run tests in mdit-py-plugins and see everything is OK there as well.

> just have to think if there could be any side effects, e.g. to plugins

There could be, if a plugin tries to mutate an item in the `attrs` list. I hope plugins wouldn't do this though, as the `Token` class provides public API for modifying `attrs`...

PR #98, opened 2020-12-16
👌 IMPROVE: Simplify `normalize_uri`

hukkinj1:


markdown_it/common/normalize_url.py
```diff
@@ -54,7 +54,7 @@ def unescape_string(s):
 
 
 def normalize_uri(uri):
-    return quote(uri.encode("utf-8"), safe=str("/@:+?=&()%#*,"))
+    return quote(uri, safe="/@:+?=&()%#*,")
```
chrisjsewell, line 57: so why exactly does this work? does quote accept bytes or strings, and why wrap the safe text in `str`?

chrisjsewell (changes requested):
the annotation conflict needs fixing ta

markdown_it/common/normalize_url.py
hukkinj1: > so why exactly does this work? does quote accept bytes or strings

Yeah `quote()` can take a string and do `utf-8` encoding internally. If we for some reason wanted to pass in `bytes`, then there is `quote_from_bytes` for that.

> and why wrap the safe text in str

No reason really, this PR removes the wrapping.

hukkinj1:
Fixed the conflict

PR #99, opened 2020-12-18
📚 DOCS: Fix docstring

tsutsu3:
before:
before

after:
after

### DIFF
--- a/markdown_it/main.py
+++ b/markdown_it/main.py
@@ -133,11 +133,10 @@ class MarkdownIt:
         containing rules with given names. If rule not found, and `ignoreInvalid`
         not set - throws exception.
 
-        ##### Example
+        Example::
 
-        ```python
-        md = MarkdownIt()..enable(['sub', 'sup']).disable('smartquotes')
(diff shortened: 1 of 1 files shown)


tsutsu3:
Ref: https://github.com/actions/setup-python/issues/171

hukkinj1:
Maybe change `pypy3` to `pypy-3.6` or `pypy-3.7` (or both) in CI as suggested in https://github.com/actions/setup-python/issues/171#issuecomment-747917974 ? That seems like the better syntax to me. It wasn't used originally here because it didn't exist yet at the time.

tsutsu3:
It has been fixed to work with  `pypy3`

tsutsu3:
> Ref: [actions/setup-python#171](https://github.com/actions/setup-python/issues/171)

It has been fixed to work with  `pypy3`

chrisjsewell:
cheers, note it won't actually show in the documentation site until https://github.com/executablebooks/markdown-it-py/pull/101 is fixed

PR #101, opened 2020-12-18
🔧 MAINTAIN: Update ReadTheDocs build

chrisjsewell:


hukkinj1:
One idea comes to mind that could be useful if the new pip dependency resolver continues to use too many resources or be slow: Use Poetry for packaging and dependencies, and do `poetry export` to dump a resolved `requirements.txt` for RTD to `pip install`.

chrisjsewell:
yeh I figured out the issue I think, that both jupytext, then myst-nb need updating to use the new version here:

- https://github.com/readthedocs/readthedocs.org/issues/7768#issuecomment-748405360
- https://github.com/mwouts/jupytext/pull/692

PR #102, opened 2020-12-19
👌 IMPROVE: Make `Token.attrs` a List[Tuple[str, Any]]

hukkinj1:


hukkinj1:
These changes do not break compatibility with `mdit-py-plugins`, but tests will need to be updated there.

hukkinj1:
https://github.com/executablebooks/mdit-py-plugins/pull/6 will fix `mdit-py-plugins` tests if this is merged.

markdown_it/renderer.py
```diff
@@ -235,16 +235,11 @@ def fence(self, tokens, idx, options, env):
         # May be, one day we will add .clone() for token and simplify this part, but
         # now we prefer to keep things local.
         if info:
-            i = token.attrIndex("class")
             tmpAttrs = token.attrs[:] if token.attrs else []
 
-            if i < 0:
-                tmpAttrs.append(["class", options.langPrefix + langName])
-            else:
-                tmpAttrs[i][1] += " " + options.langPrefix + langName
```
hukkinj1, line 244: This line actually seems to have a bug that we're fixing by using the immutable tuple :smile: Mutating `tmpAttrs[i]` here will also mutate `token.attrs[i]` because `tmpAttrs` is a copy, not a deep copy, of `token.attrs`.

hukkinj1: Here's the fix in JS markdown-it https://github.com/markdown-it/markdown-it/commit/c9949dd585315990689a6bb361388ea30c94638a

If we don't end up merging this PR, we should do the same change.

hukkinj1:
In fact, if we want to make `Token.attrs` even better, we could use a `Dict[str, Any]` as the underlying structure. Why `markdown-it` doesn't use an object is exlained here https://github.com/markdown-it/markdown-it/issues/142#issuecomment-133042149 : 

> Because ordering makes sence for testing, and no needs to care about hacks like "proto" name and so on.

A Python 3.6 `OrderedDict` or Python 3.7+ `dict` doesn't suffer from any of the issues why the custom list structure was used for a mapping. Using a `dict` would likely result in a small performance increase.

markdown_it/renderer.py
hukkinj1: This fix is now in merged in `master` of markdown-it-py btw

chrisjsewell:
Superseded by #144

PR #105, opened 2020-12-21
🧪 TEST: Enable typographer to fix skipped

tsutsu3:
Nothing to do with #8

`typographer` was enabled in the [js version](https://github.com/markdown-it/markdown-it/blob/master/test/markdown-it.js#L14). 

`---` will be converted to a `—`(&mdash).


chrisjsewell:
one test down cheers!

PR #106, opened 2020-12-22
👌 IMPROVE: Add more typing

hukkinj1:


markdown_it/rules_core/inline.py
```diff
@@ -5,4 +5,5 @@ def inline(state: StateCore) -> None:
     """Parse inlines"""
     for token in state.tokens:
         if token.type == "inline":
+            assert token.children is not None
```
chrisjsewell, line 8: I think this is more fault tolerant, e.g. if a plugin does not specifically set `token.children = []` (the default is `None`)

```suggestion
            if token.children is None:
                token.children = []
```

hukkinj1: :+1: Note that we have this same assertion in the codebase in a few other places too, so might want to change them too. IMO this assertion should be fine, though. The token type is "inline" so children should never be `None` in that case. Unfortunately this information is quite dynamic so our typing doesn't know about it (hence the assertion is required).

JS markdown-it for instance will crash hard if inline token children is ever `null` (e.g. this line https://github.com/markdown-it/markdown-it/blob/7b8969ce5cb2edc54f2c1aa39a85a3a08076337d/lib/renderer.js#L271)

PR #108, opened 2020-12-25
👌 IMPROVE: Ensure len(state.src) == len(state.srcCharCode)

geebos:
```python
def normalize(state: StateCore) -> None:

    # Normalize newlines
    string = NEWLINES_RE.sub("\n", state.src)

    # Replace NULL characters
    string = NULL_RE.sub("\uFFFD", string)

    state.src = string
```

When updating state.src, state.srcCharCode should be updated synchronously.

### DIFF
--- a/markdown_it/ruler.py
+++ b/markdown_it/ruler.py
@@ -35,6 +35,9 @@ class StateBase:
         self._src = value
         self.srcCharCode = [ord(c) for c in self.src] if self.src is not None else []
 
+    def set_underline_src(self, src):
+        self._src = src
+
 
 # The first positional arg is always a subtype of `StateBase`. Other
 # arguments may or may not exist, based on the rule's type (block,
--- a/markdown_it/ruler.py
+++ b/markdown_it/ruler.py
@@ -25,6 +25,19 @@ class StateBase:
         self.env = env
         self.md = md
 
+    @property
+    def src(self):
+        return self._src
+
+    @src.setter
+    def src(self, value):
+        self._src = value
+        self._srcCharCode = [ord(c) for c in self.src]
+
+    @property
+    def srcCharCode(self):
+        # To avoid assigning srcCharCode
+        return self._srcCharCode
 
 # The first positional arg is always a subtype of `StateBase`. Other
 # arguments may or may not exist, based on the rule's type (block,
--- a/markdown_it/ruler.py
+++ b/markdown_it/ruler.py
@@ -34,10 +34,6 @@ class StateBase:
         self._src = value
         self._srcCharCode = [ord(c) for c in self.src]
 
-    @property
-    def srcCharCode(self):
-        # To avoid assigning srcCharCode
-        return self._srcCharCode
 
 # The first positional arg is always a subtype of `StateBase`. Other
 # arguments may or may not exist, based on the rule's type (block,
--- a/markdown_it/ruler.py
+++ b/markdown_it/ruler.py
@@ -21,7 +21,7 @@ import attr
 
 class StateBase:
     def __init__(self, src: str, md, env):
-        self.srcCharCode = []
+        self.srcCharCode: List[int] = []
         self.src = src
         self.env = env
         self.md = md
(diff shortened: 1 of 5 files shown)


markdown_it/rules_core/state_core.py
```diff
@@ -12,3 +12,7 @@ def __init__(self, src: str, md, env, tokens=None):
         self.env = env
         self.tokens: List[Token] = tokens or []
         self.inlineMode = False
+
+    def set_src(self, src):
+        self.src = src
+        self.srcCharCode = [ord(c) for c in src] if src is not None else []
```
hukkinj1, line 18:
```suggestion
    @property
    def src(self):
        return self._src

    @src.setter
    def src(self, value):
        self._src = value
        self.srcCharCode = [ord(c) for c in self.src] if self.src is not None else []
```

markdown_it/rules_core/normalize.py
```diff
@@ -17,4 +17,4 @@ def normalize(state: StateCore) -> None:
     # Replace NULL characters
     string = NULL_RE.sub("\uFFFD", string)
 
-    state.src = string
+    state.set_src(string)
```
hukkinj1, line 20:
```suggestion
    state.src = string
```

hukkinj1 (commented):
Nice! I made suggestions that use the `@property` decorator to keep API identical to JS markdown-it.

With those changes I believe we can also remove line 10 from the `__init__` that sets `self.srcCharCode`.

hukkinj1:
Also, might make sense to add the `@property` to `StateBase` base class only, to get identical behavior in all `StateBase` subclasses.

geebos:
> Also, might make sense to add the `@property` to `StateBase` base class only, to get identical behavior in all `StateBase` subclasses.

Nice suggestions! I have added the `@property` to `StateBase`. In addition, I have added `srcCharCode` to `@property` to avoid assigning srcCharCode from outer codes.

markdown_it/ruler.py
```diff
@@ -25,6 +25,19 @@ def __init__(self, src: str, md, env):
...
+
+    @src.setter
+    def src(self, value):
+        self._src = value
+        self._srcCharCode = [ord(c) for c in self.src]
+
+    @property
+    def srcCharCode(self):
+        # To avoid assigning srcCharCode
+        return self._srcCharCode
```
hukkinj1, line 40:
```suggestion
```

I think we have to remove this property and keep `srcCharCode` writeable. We write it directly in `StateBlock.__init__`. It seems to me like a performance optimization, but @chrisjsewell knows better.

hukkinj1 (changes requested):
We should be able to remove the line
```python
self.srcCharCode = [ord(c) for c in src] if src is not None else []
```
in `StateInline.__init__`.

In `StateBlock.__init__` we can make the lines
```python
        self.src = src
        if srcCharCode is not None:
            self.srcCharCode = srcCharCode
        else:
            self.srcCharCode = [ord(c) for c in src] if src is not None else []
```
something like
```python
        if srcCharCode is not None:
            self._src = src
            self.srcCharCode = srcCharCode
        else:
            self.src = src
```

This will keep the (presumably) performance improvement in place.

hukkinj1:
Looking good @geebos ! Did you see this comment https://github.com/executablebooks/markdown-it-py/pull/108#pullrequestreview-558823485 . You think those two changes can be used here too?

geebos:
> Looking good @geebos ! Did you see this comment [#108 (review)](https://github.com/executablebooks/markdown-it-py/pull/108#pullrequestreview-558823485) . You think those two changes can be used here too?

Thanks for reminding. I have added a `set_underline_src` method in `StateBase`, because we cant set `_src` directly in `StateBlock`.

hukkinj1:
> we cant set _src directly in StateBlock.

What makes you think so? I believe this should work just fine as long as there's only one preceding underscore (`_src`), not two (`__src`). In case it's mypy that's giving you trouble, you can type annotate `self._src` in `StateBase` after the class declaration like
```python
class StateBase:
    _src: str
```

geebos:
You are right. I confuse them.

hukkinj1:
Nice, this looks very good to me!

chrisjsewell:
thanks!

PR #109, opened 2020-12-26
🐛 FIX: port markdown-it v11.0.1

hukkinj1:
Implement the diff https://github.com/markdown-it/markdown-it/compare/f798bea9623277bbf89b9621cf7fb283c693fcab...331ae117e09115366db12b517ca526b1b7fee1e8 to catch up with the JS project a bit.

PR #110, opened 2020-12-27
Catch up with markdown-it v12.0.0

hukkinj1:
A continuation on https://github.com/executablebooks/markdown-it-py/pull/109 (based on the branch there). 

https://github.com/executablebooks/markdown-it-py/pull/109 can be merged first and then this one (to make review easier), or this can be merged directly.

### DIFF
--- a/tests/test_port/fixtures/tables.md
+++ b/tests/test_port/fixtures/tables.md
@@ -273,106 +273,6 @@ bar|bar
 .
 
 
-Should be terminated via row without "|" symbol:
-.
-foo|foo
----|---
-paragraph
-.
-<table>
-<thead>
-<tr>
-<th>foo</th>
-<th>foo</th>
-</tr>
-</thead>
-<tbody></tbody>
-</table>
-<p>paragraph</p>
-.
-
-
-Delimiter escaping:
-.
-| Heading 1 \\\\| Heading 2
-| --------- | ---------
-| Cell\|1\|| Cell\|2
-\| Cell\\\|3 \\| Cell\|4
-.
-<table>
-<thead>
-<tr>
-<th>Heading 1 \\</th>
-<th>Heading 2</th>
-</tr>
-</thead>
-<tbody>
-<tr>
-<td>Cell|1|</td>
-<td>Cell|2</td>
-</tr>
-<tr>
-<td>| Cell\|3 \</td>
-<td>Cell|4</td>
-</tr>
-</tbody>
-</table>
-.
-
-Pipes inside backticks don't split cells:
-.
-| Heading 1 | Heading 2
-| --------- | ---------
-| Cell 1 | Cell 2
-| `Cell|3` | Cell 4
-.
-<table>
-<thead>
-<tr>
-<th>Heading 1</th>
-<th>Heading 2</th>
-</tr>
-</thead>
-<tbody>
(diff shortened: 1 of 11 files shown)


hukkinj1:
## [12.0.4] - 2020-12-20
### Fixed
- Fix crash introduced in `12.0.3` when processing strikethrough (`~~`) and similar plugins, #742.
- Avoid fenced token mutation, #745.


## [12.0.3] - 2020-12-07
### Fixed
- `[](<foo<bar>)` is no longer a valid link.
- `[](url (xxx())` is no longer a valid link.
- `[](url\ xxx)` is no longer a valid link.
- Fix performance issues when parsing links (#732, #734), backticks, (#733, #736),
  emphases (#735), and autolinks (#737).
- Allow newline in `<? ... ?>` in an inline context.
- Allow `<meta>` html tag to appear in an inline context.


## [12.0.2] - 2020-10-23
### Fixed
- Three pipes (`|\n|\n|`) are no longer rendered as a table with no columns, #724.


## [12.0.1] - 2020-10-19
### Fixed
- Fix tables inside lists indented with tabs, #721.


## [12.0.0] - 2020-10-14
### Changed
- Added 3rd argument to `highlight(code, lang, attrs)`, #626.
- Rewrite tables according to latest GFM spec, #697.

### Fixed
- Fix mappings for table rows (amended fix made in 11.0.1), #705.
- `%25` is no longer decoded in beautified urls, #720.

chrisjsewell:
I think maybe split this up into a few defined commits, e.g. one for links, one for the table etc?

For tables what is the actual difference? are there table syntaxes that were not allowed before or are allowed now? Is there any tests we can add to highlight the difference?

chrisjsewell:
> is there any tests we can add to highlight the difference?

ah oops, I missed the changes to tables.md (because the diff was collapsed)

tests/test_port/fixtures/tables.md
```diff
@@ -273,106 +273,6 @@ bar|bar
 .
 
 
-Should be terminated via row without "|" symbol:
```
chrisjsewell, line 276: I think it would be best to leave these tests in and, if necessary, change the expected outcomes and maybe the heading.
Just so it's clear what the new behaviour is

hukkinj1: :+1: added back the tests

hukkinj1:
> I think maybe split this up into a few defined commits, e.g. one for links, one for the table etc?

This might be nice to have, but at least I find it hard to motivate myself to do this :smile: I don't mind at all if you take the changes and redo the commit history.

> For tables what is the actual difference? are there table syntaxes that were not allowed before or are allowed now? Is there any tests we can add to highlight the difference?

The test cases I re-added show some changes. I think it's a bit hard to argue more about the changes, since previously there was no formal spec that was followed.

chrisjsewell:
> I don't mind at all if you take the changes and redo the commit history.

@hukkinj1 with any luck this should have now been fully covered by #140
If you want to check and close this

hukkinj1:
Nice, I merged master to this and as a result there's still a few changes left here that seem like they belong in master.

chrisjsewell:
From the merge conflicts I noticed a few minor errors that I'm just about to commit

PR #117, opened 2021-01-06
DRAFT: Make nest_tokens preserve unmodified tokens

hukkinj1:
Working on mdformat I bumped into some problems where having access to the Markdown in a tree structure might simplify and prevent some hacks (doing https://github.com/executablebooks/mdformat/pull/85). I looked into `nest_tokens` and noticed it modifies and doesn't fully respect typing of `Token.children`, adding `NestedTokens` objects into a `List[Token]`. It could be useful for rendering (and perhaps other) purposes to keep the original token.

I made some tweaks that slightly change the meaning of `NestedTokens` making it more of a container of `Token`s than a replacement for them. This change makes `nest_token` respect typing of `Token.attrs`, makes rendering possibly simpler since the original tokens are kept, and probably makes it simpler to reconstruct the linear token stream.

This a draft to raise discussion. I'm not sure if I like this :smile: not sure if @chrisjsewell likes this, and some docstrings, names etc. might need to be changed before merging. I'm also not aware of how widely `NestedTokens` and `nest_tokens` are used and how destructive a change like this would be.

### DIFF
--- a/markdown_it/token.py
+++ b/markdown_it/token.py
@@ -1,4 +1,4 @@
-from typing import List, Optional, Union
+from typing import List, Optional
 
 import attr
 
@@ -117,8 +117,8 @@ class NestedTokens:
     """
 
     opening: Token = attr.ib()
-    closing: Optional[Token] = attr.ib()
-    children: List[Union[Token, "NestedTokens"]] = attr.ib(factory=list)
+    closing: Optional[Token] = attr.ib(default=None)
+    children: List["NestedTokens"] = attr.ib(factory=list)
 
     def __getattr__(self, name):
         return getattr(self.opening, name)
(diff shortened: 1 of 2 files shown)


chrisjsewell:
Errr, I'll have a look, but would be very hesitant to change anything unless it is really needed, since it is basically the backbone of https://github.com/executablebooks/MyST-Parser, and hence https://github.com/executablebooks/jupyter-book.
Given this I can say it is very widely used and also very well (indirectly) tested

chrisjsewell:
This is where it is used if your interested: https://github.com/executablebooks/MyST-Parser/blob/396bde76823b675bb9a55f414bb391a2f24744bc/myst_parser/docutils_renderer.py#L116

hukkinj1:
Yeah if thats the case, no worries. This is something thats really simple to reimplement in a different manner in a downstream project anyways.

chrisjsewell:
Yeh I don't want to give myself anymore work to do right now 😂

PR #120, opened 2021-01-12
👌 IMPROVE: Use __all__ to signal re-exports

hukkin:


hukkin:
@chrisjsewell To clarify why this is useful: this allows downstream projects (and markdown-it-py itself) that use the distributed type data to set `implicit_reexport = False` mypy flag. This flag will protect from accidental bad imports by making them mypy errors. Here's examples of bad imports that unfortunately do work at runtime but are not public API and should never be used:
```python
from markdown_it.token import attr
from markdown_it.tree import Token
```

If you dislike setting the `__all__` attribute, an alternative way to express explicit re-export that mypy understands is an import using "`as`":
```python
from .state_core import StateCore as StateCore
```

chrisjsewell:
cheers, I think this is fine, one nitpick would be I think the `__all__` should be underneath the imports, unless you no of a reason why not?
and then can you also look at fixing the documentation build fail

hukkin:
> one nitpick would be I think the __all__ should be underneath the imports, unless you no of a reason why not?

This is simply what [PEP8](https://www.python.org/dev/peps/pep-0008/#module-level-dunder-names) tells us to do so following that.

Hmm, I'll have a look at the docs failure in a short while.

hukkin:
The issue seems to be as follows:

```python
>>> from markdown_it.rules_block.blockquote import blockquote
>>> repr(blockquote)
Out[3]: '<function blockquote at 0x7f7ebe68aaf0>'
>>> from markdown_it.rules_block import blockquote
>>> repr(blockquote)
Out[5]: '<function blockquote at 0x7f7ebe68aaf0>'
```

The name `markdown_it.rules_block.blockquote` is overloaded. It is a module to import from and also an importable function. Sphinx autodocs do not seem to like this very much.

I suggest renaming the `markdown_it.rules_block.blockquote` as `markdown_it.rules_block._blockquote` (i.e. with an underscore). That fixes the error and is in line with what's importable in JS markdown-it.

Another way is to just ignore the Sphinx warnings.

What do you think?

chrisjsewell:
> Hmm, I'll have a look at the docs failure in a short while.

Sorted it 😄 

> I suggest renaming the `markdown_it.rules_block.blockquote` module as `markdown_it.rules_block._blockquote` (i.e. with an underscore). That fixes the error and is in line with what's importable in JS markdown-it.

This is possibly the better solution eventually, but err I'd want to think about it some more, and so don't want to hold up this PR

PR #124, opened 2021-02-15
🐛 FIX: cli.parse: return 0 for non-error

hukkinj1:
Based on @westurner's https://github.com/executablebooks/markdown-it-py/pull/55 with two changes:

- Fix CI
- Fix error code when running `python markdown_it/cli/parse.py`


westurner:
Thanks!

Issue #121, opened 2021-01-20
Add an abstract renderer class


chrisjsewell:
Mainly to aid in typing

PR #126, opened 2021-02-22
👌 IMPROVE: Add `RendererProtocol` typing

hukkinj1:
Closes https://github.com/executablebooks/markdown-it-py/issues/121


chrisjsewell (changes requested):

markdown_it/renderer.py
```diff
@@ -6,14 +6,34 @@ class Renderer
...
+)
 
 from .common.utils import unescapeAll, escapeHtml
 from .token import Token
 from .utils import OptionsDict
 
+if sys.version_info < (3, 8):
+    from typing_extensions import Protocol
+else:
+    from typing import Protocol
```
chrisjsewell, line 24:
```suggestion
try:
	from typing import Protocol
except ImportError:
    from typing_extensions import Protocol    
```

I've literally just has an issue where `sys.version_info` was not reliable

chrisjsewell (changes requested):

markdown_it/renderer.py
hukkinj1: Changed. Please let me know what the issue was. I'm very interested.

markdown_it/renderer.py
```diff
@@ -6,14 +6,33 @@ class Renderer
...
+    from typing_extensions import Protocol  # type: ignore[misc]
 
-class RendererHTML:
+
+class RendererProtocol(Protocol):
+    __output__: ClassVar[str]
+
+    def render(
+        self, tokens: Sequence[Token], options: OptionsDict, env: MutableMapping
+    ) -> str:
```
chrisjsewell, line 31:
```suggestion
    ) -> Any:
```

for example `DocutilsRenderer.render` returns a `docutils.nodes.document`
(the `MarkdownIt.render` method also already returns `Any`)

markdown_it/renderer.py
chrisjsewell: It actually turned out to be an issue with importing of `importlib_metadata` instead of `importlib-metadata` in a Conda environment 😒 (see https://github.com/python/importlib_metadata/issues/308 and executablebooks/MyST-NB#321), which actually illustrates my "worry" in https://github.com/executablebooks/markdown-it-py/pull/154#issuecomment-811740183

But anyhow, I think this approach is still slightly "safer"

markdown_it/renderer.py
hukkinj1: :+1: committed

chrisjsewell: guess you need to also import Any

markdown_it/renderer.py
```diff
@@ -6,14 +6,33 @@ class Renderer
 rules if you create plugin and adds new token types.
 """
 import inspect
-from typing import MutableMapping, Optional, Sequence
+from typing import (
+    ClassVar,
```
chrisjsewell, line 10:
```suggestion
    Any,
    ClassVar,
```

markdown_it/renderer.py
hukkinj1: Ah, so conda not doing the name normalization if I understand correctly? Good to know, thanks! I don't use Conda so no need to change any of my personal projects :+1:

chrisjsewell: yeh seems to be the case 😬

chrisjsewell: despite that it is generally very good; particularly in resolving dependencies and/or if you are using non-python dependencies

chrisjsewell:
ok thats it; v1 is getting released (after doing another beta test). You can propose other changes, but I will probably ignore you 😜

hukkinj1:
The only thing I think would still be great is the [`AttrDict` / `entities` PR](https://github.com/executablebooks/markdown-it-py/pull/153) and [changing `entities` import path](https://github.com/executablebooks/markdown-it-py/pull/153#issuecomment-811047566) for consistency with upstream (and removal of stutter).

These are breaking changes so will not be fun doing after v1.

Although breaking, these aren't that major changes so feel free to ignore :smile:

chrisjsewell:
> These are breaking changes so will not be fun doing after v1.

Well I've already been enforcing deprecation path ways for breaking changes since 0.6 (like plugin removal), so I'd still have looked to do that even if it was not v1.

and yeh its just such a minimally used thing to let it hold up the release

PR #129, opened 2021-02-25
✨ NEW: Add SyntaxTreeNode

hukkinj1:
Context: read the thread here https://github.com/executablebooks/mdformat/pull/132#discussion_r582360434

### DIFF
--- a/markdown_it/tree.py
+++ b/markdown_it/tree.py
@@ -12,7 +12,8 @@ class SyntaxTreeNode:
     """A Markdown syntax tree node.
 
     A class that can be used to construct a tree representation of a linear
-    `markdown-it-py` token stream.
+    `markdown-it-py` token stream. Use `SyntaxTreeNode.from_tokens` to
+    initialize instead of the `__init__` method.
 
     Each node in the tree represents either:
       - root of the Markdown document
@@ -45,11 +46,17 @@ class SyntaxTreeNode:
 
     @staticmethod
     def from_tokens(tokens: Sequence[Token]) -> "SyntaxTreeNode":
+        """Instantiate a `SyntaxTreeNode` from a token stream.
+
+        This is the standard method for instantiating `SyntaxTreeNode`.
+        """
         root = SyntaxTreeNode()
         root._set_children_from_tokens(tokens)
         return root
 
     def to_tokens(self) -> List[Token]:
+        """Recover the linear token stream."""
+
         def recursive_collect_tokens(
             node: "SyntaxTreeNode", token_list: List[Token]
         ) -> None:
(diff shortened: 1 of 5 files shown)


chrisjsewell:
Thanks firstly can you move this to a separate file: tree.py and also note in the module that this is not part of upstream markdown-it
Then also add a deprecation warning inside nest_tokens

hukkinj1:
@chrisjsewell :+1: done

chrisjsewell:
thanks, next can you add docstrings for all the methods and were we going to have `is_nested` and `is_root` properties?

markdown_it/tree.py
```diff
@@ -0,0 +1,211 @@
...
+This module is not part of upstream JavaScript markdown-it.
+"""
+from typing import NamedTuple, Sequence, Tuple, Dict, List, Optional, Any
+
+from .token import Token
+from .utils import _removesuffix
+
+
+class SyntaxTreeNode:
+    """A Markdown syntax tree node.
```
chrisjsewell, line 12: perhaps also add a note here, that it should be initialised by `SyntaxTreeNode.from_tokens`

hukkinj1:
> thanks, next can you add docstrings for all the methods

Sure thing.

> and were we going to have is_nested and is_root properties?

What do you think? I realized I don't have much need for these.

`is_root` basically equals `not self.parent` and that can be useful when traversing towards root and checking if you're already there. But in that context `not self.parent` feels more intuitive, because if you're not there yet, what you do next is `self.parent.parent`...

`is_nested` I never need in mdformat. That would be equal to `bool(self.nester_tokens)` which isn't overly onerous to do. And if the use case for `is_nested` is checking before accesing `self.nester_tokens` then `bool(self.nester_tokens)` could be more intuitive?

So basically just didn't add these because they weren't useful to me, but have no problem with adding them.

chrisjsewell:
yeh I think at least add the `is_nested`

hukkinj1:
We could enforce `__init__` privateness (in favor of `from_tokens`) by something like adding a private mandatory kwarg:

```python
def __init__(self, __internal_call=False):
    if not __internal_call:
        raise Exception("this is private API")
```

just an idea, not sure if a good one :smile:

chrisjsewell:
> just an idea, not sure if a good one

yeh not necessarily a bad idea lol, but probably overkill

hukkinj1:
Added docstrings and `is_nested`. Didn't add docstrings for the `@property`ies that are just a passthrough from a `Token`. Do you think we need? There is a comment that kind of explains them as a whole.

chrisjsewell:
> dn't add docstrings for the @propertyies that are just a passthrough from a Token. Do you think we need? 

I would say yes (basically copied from the comments in `Token`), because they are helpful when e.g. working in editors like VS Code where you get the description when you hover over them

hukkinj1:
Yep, makes sense. Copied those from `Token` class.

chrisjsewell:
lastly, can you add a few extra bits to the test module, to increase the test coverage

hukkinj1:
Added. Tell me if you think there's something else in particular that you want tested.

chrisjsewell:
nah thats great cheers

PR #130, opened 2021-02-25
👌 IMPROVE: Make `SyntaxTreeNode` subclassable

hukkinj1:
Realized I can not subclass and add methods to the tree class, because it always instantiates the base class. This fixes it


chrisjsewell:
👍

PR #131, opened 2021-02-26
👌 IMPROVE: Make SyntaxTreeNode type annotations work when subclassing

hukkinj1:
~~This PR makes `SyntaxTreeNode.children` and `SyntaxTreeNode.parent` read-only properties, as it seems that's by far the least complicated way of making type annotations work expectedly when extending the class. Those properties shouldn't be rewritten anyways so I guess this is not a bad thing~~
Edit: This PR simply makes `SyntaxTreeNode.children` and `SyntaxTreeNode.parent` `@property`s so that they can be type annotated in a way where the getter return type `TypeVar` is automatically bound to type of `self`.


hukkinj1:
@chrisjsewell would be great if we could sort this out before the next release. Making the two attributes is a breaking change, but if this is dealt with before the next release, then it will never be release breaking change.

There's two commits in this PR. The first commit is an alternative self-contained working solution, and doesn't require making the attributes read-only. It however makes the class a `Generic` that you'd always have to subclass like
```python
class MyTree(SyntaxTreeNodeBase["MyTree"]):
    pass
```
or alternatively always annotate or instantiate like
```python
tree = MyTree["MyTree"]()
```
At least I couldn't find any way to to make the type checker automatically bind the `TypeVar` to the subclass type so that this wouldn't be needed.

Soo, that's very awkward so I think I much prefer the state after the second commit.

hukkinj1:
If you're wondering why this has to be so complicated in the first place, haha, it's because many interfaces (method input or outputs, public class attrs) in this class reference the type of `self`, and that type changes when subclassing. This isn't too common so that's why type annotations are not always so hard.

markdown_it/tree.py
```diff
@@ -2,12 +2,30 @@
...
 from .token import Token
 from .utils import _removesuffix
 
 
+class _NesterTokens(NamedTuple):
+    opening: Token
+    closing: Token
+
+
+_T = TypeVar("_T", bound="SyntaxTreeNode")
```
chrisjsewell, line 26: why `_T`? it would be better from a readability perspective to use something like `NodeType`

markdown_it/tree.py
```diff
@@ -35,17 +49,17 @@ def __init__(self) -> None:
...
+        self.nester_tokens: Optional[_NesterTokens] = None
 
         # Root node does not have self.parent
-        self.parent: Optional["SyntaxTreeNode"] = None
+        self.parent = None
 
         # Empty list unless a non-empty container, or unnested token that has
         # children (i.e. inline or img)
-        self.children: List["SyntaxTreeNode"] = []
+        self.children = []
```
chrisjsewell, line 59: should this not be `self._children`?

markdown_it/tree.py
```diff
@@ -35,17 +49,17 @@ def __init__(self) -> None:
         self.token: Optional[Token] = None
 
         # Only containers have nester tokens
-        self.nester_tokens: Optional[SyntaxTreeNode._NesterTokens] = None
+        self.nester_tokens: Optional[_NesterTokens] = None
 
         # Root node does not have self.parent
-        self.parent: Optional["SyntaxTreeNode"] = None
+        self.parent = None
```
chrisjsewell, line 55: should this not be `self._parent`

markdown_it/tree.py
```diff
@@ -35,17 +49,17 @@ def __init__(self) -> None:
...
 
         # Empty list unless a non-empty container, or unnested token that has
         # children (i.e. inline or img)
-        self.children: List["SyntaxTreeNode"] = []
+        self.children = []
 
     @classmethod
-    def from_tokens(cls, tokens: Sequence[Token]) -> "SyntaxTreeNode":
+    def from_tokens(cls: Type[_T], tokens: Sequence[Token]) -> _T:
```
chrisjsewell, line 62: do `cls` and `self` really need to be annotated?

chrisjsewell (changes requested):

markdown_it/tree.py
hukkinj1: It seems like its a convention to name `TypeVar`s as `T` (or if private, `_T`). I think this is because its short, and because the same `TypeVar` could be reused in various contexts having a different meaning in each. Here setting the `bound` sort of limits any reasonable reusability though.

That being said, I don't oppose renaming at all if you think thats better :smile:

markdown_it/tree.py
hukkinj1: There really is no difference, but I think using the public API is less brittle in case someone ever changes the `@property` setter

markdown_it/tree.py
hukkinj1: same as self._children :smile:

markdown_it/tree.py
chrisjsewell: yes I think so thanks

markdown_it/tree.py
chrisjsewell: I think all internal variables should be set in `__init__`. TBH I'm suprised flake8 does not fail this, pylint definitely does

markdown_it/tree.py
hukkinj1: Yep, the type checker binds a meaning to `_T` based on context. Here it gets the context from `cls` (it knows its type) so it deduces that the return type is the same as that.

E.g. if you call `node = MyTreeNode.from_tokens()`, the type checker then knows that `node` is `MyTreeNode` and if the type is something different (another subclass) the type checker binds the `TypeVar` to that.

chrisjsewell: ok cheers

hukkinj1: To clarify, if you dont annnotate `cls`, the type checker doesn't have any context useful to binding the `TypeVar` so the return type will be `Any`.

markdown_it/tree.py
hukkinj1: The internal vars are set in init, but via the @property, flake8 understands this.

markdown_it/tree.py
hukkinj1: changed to `_NodeType`

markdown_it/tree.py
hukkinj1: Reluctantly made this change :smile:

hukkinj1:
@chrisjsewell I added the necessary `@overload`s to type annotate `__getitem__` properly in https://github.com/executablebooks/markdown-it-py/pull/131/commits/c0b698af9318ccc8c8c4f3b91a1f5a0a2d4739a4

chrisjsewell (approved):

PR #132, opened 2021-02-28
👌 IMPROVE: Use regular __init__ to create SyntaxTreeNodes

hukkinj1:
Based on https://github.com/executablebooks/markdown-it-py/pull/131

- Always use `__init__` as constructor, remove `from_tokens`
-  Allow constructing trees that don't have a document root, using an inline token or nested tokens as input.


chrisjsewell:
note this will aslo require a change to the documentation: https://markdown-it-py.readthedocs.io/en/latest/using.html#creating-a-syntax-tree

hukkinj1:
Fixed the `from_tokens` reference in docs.

chrisjsewell:
@hukkinj1 can you resolve the conflicts

hukkinj1:
:+1: Rebased on master

markdown_it/tree.py
```diff
@@ -43,10 +41,12 @@ class SyntaxTreeNode:
...
 
-    def __init__(self) -> None:
-        """Initialize a root node with no children.
+    def __init__(
+        self, tokens: Sequence[Token] = (), *, create_root: bool = True
+    ) -> None:
+        """Initialize a `SyntaxTreeNode` from a token stream.
 
-        You probably need `SyntaxTreeNode.from_tokens` instead.
+        If `create_root` is True, crete a root node for the document.
```
chrisjsewell, line 49:
```suggestion
        If `create_root` is True, create a root node for the document.
```

markdown_it/tree.py
```diff
@@ -61,6 +61,25 @@ def __init__(self) -> None:
...
+            return
+
+        if not tokens:
+            raise ValueError(
+                "Can only create root from empty token sequence."
+                " Set `create_root=True`."
+            )
+        elif len(tokens) == 1:
+            inline_token = tokens[0]
+            assert not inline_token.nesting
```
chrisjsewell, line 75: maybe add an assertion message here, why this is not allowed

markdown_it/tree.py
```diff
@@ -211,27 +211,22 @@ def _set_children_from_tokens(self, tokens: Sequence[Token]) -> None:
...
-                child = self._make_child(token=token)
-                if token.children:
-                    child._set_children_from_tokens(token.children)
+            if not token.nesting:
+                self._add_child([token])
                 continue
-
-            assert token.nesting == 1
+            if token.nesting != 1:
+                raise ValueError("Invalid token nesting")
```
chrisjsewell, line 221: it would be nice here to identify where in the token stream the issue is.
Maybe at the start of this function set `length = len(tokens)`, then here do `index = length - len(reversed_tokens)` and add this to the exception message

markdown_it/tree.py
```diff
@@ -211,27 +211,22 @@ def _set_children_from_tokens(self, tokens: Sequence[Token]) -> None:
...
             nested_tokens = [token]
             nesting = 1
-            while reversed_tokens and nesting != 0:
+            while reversed_tokens and nesting:
                 token = reversed_tokens.pop()
                 nested_tokens.append(token)
                 nesting += token.nesting
-            if nesting != 0:
+            if nesting:
                 raise ValueError(f"unclosed tokens starting {nested_tokens[0]}")
```
chrisjsewell, line 230: maybe same here as above for noting the index of the problematic token in the token stream

chrisjsewell (changes requested):
just a few minor suggestions

markdown_it/tree.py
hukkinj1: Fixed

markdown_it/tree.py
hukkinj1: Fixed

markdown_it/tree.py
hukkinj1: Note that due to the nested nature of the tree, there's a very good chance `tokens` won't be the same stream that the programmer input to `SyntaxTreeNode` init but rather a nested sub-stream. In this case showing the index could potentially be more confusing than helpful. I'm also a bit hesitant to add logic just for the sake of an exception message that, in software working expectedly, should never be raised.

Can add this if you still think its a good idea.

chrisjsewell: ah yeh fair I think, if only for myst-parser, I want to have a new exception (i.e. that inherits `ValueError`) that stores the token, so I can report maybe the line number (from token.map) or whole token, but I can add that in a separate PR

PR #136, opened 2021-03-03
👌 IMPROVE: MarkdownIt config and documentation

chrisjsewell:
Add additional configuration presets,
allow for options tp be overriden in the `MarkdownIt` initialisation,
Add convenience methods to `SyntaxTreeNode`

### DIFF
--- a/docs/using.md
+++ b/docs/using.md
@@ -149,7 +149,7 @@ md.render("'single quotes' (c)")
 ### Linkify
 
 The `linkify` component requires that [linkify-it-py](https://github.com/tsutsu3/linkify-it-py) be installed (e.g. *via* `pip install markdown-it-py[linkify]`).
-This allows URL links to be identified, without the need for enclosing in `<>` brackets:
+This allows URI autolinks to be identified, without the need for enclosing in `<>` brackets:
 
 ```{code-cell}
 md = MarkdownIt("commonmark", {"linkify": True})
--- a/docs/using.md
+++ b/docs/using.md
@@ -56,7 +56,7 @@ You can define this configuration *via* directly supplying a dictionary or a pre
 - `js-default`: This is the default in the JavaScript version.
   Compared to `commonmark`, it disables HTML parsing and enables the table and strikethrough components.
 - `gfm-like`: This configures the parser to approximately comply with the [GitHub Flavored Markdown specification](https://github.github.com/gfm/).
-  Compared to `commonmark`, it disables HTML parsing and enables the table, strikethrough and linkify components.
+  Compared to `commonmark`, it enables the table, strikethrough and linkify components.
   **Important**, to use this configuration you must have `linkify-it-py` installed.
 
 ```{code-cell}
@@ -279,7 +279,6 @@ node.children
 print(node[0])
 node[0].next_sibling
 ```
-
 
 ## Renderers
 
--- a/docs/using.md
+++ b/docs/using.md
@@ -28,6 +28,7 @@ then these are converted to other formats using 'renderers'.
 The simplest way to understand how text will be parsed is using:
 
 ```{code-cell}
+from pprint import pprint
 from markdown_it import MarkdownIt
 ```
 
@@ -48,8 +49,15 @@ for token in md.parse("some *text*"):
 
 The `MarkdownIt` class is instantiated with parsing configuration options,
 dictating the syntax rules and additional options for the parser and renderer.
-You can define this configuration *via* a preset name (`'zero'`, `'commonmark'` or `'default'`),
-or by directly supplying a dictionary.
+You can define this configuration *via* directly supplying a dictionary or a preset name:
+
+- `zero`: This configures the minimum components to parse text (i.e. just paragraphs and text)
+- `commonmark` (default): This configures the parser to strictly comply with the [CommonMark specification](http://spec.commonmark.org/).
+- `js-default`: This is the default in the JavaScript version.
+  Compared to `commonmark`, it disables HTML parsing and enables the table and strikethrough components.
+- `gfm-like`: This configures the parser to approximately comply with the [GitHub Flavored Markdown specification](https://github.github.com/gfm/).
+  Compared to `commonmark`, it disables HTML parsing and enables the table, strikethrough and linkify components.
+  **Important**, to use this configuration you must have `linkify-it-py` installed.
 
 ```{code-cell}
 from markdown_it.presets import zero
@@ -61,18 +69,26 @@ md = MarkdownIt("zero")
 md.options
 ```
 
+You can also override specific options:
+
 ```{code-cell}
-print(md.get_active_rules())
+md = MarkdownIt("zero", {"maxNesting": 99})
+md.options
 ```
 
 ```{code-cell}
-print(md.get_all_rules())
+pprint(md.get_active_rules())
 ```
 
 You can find all the parsing rules in the source code:
 `parser_core.py`, `parser_block.py`,
 `parser_inline.py`.
-Any of the parsing rules can be enabled/disabled, and these methods are chainable:
+
+```{code-cell}
+pprint(md.get_all_rules())
+```
+
+Any of the parsing rules can be enabled/disabled, and these methods are "chainable":
 
 ```{code-cell}
 md.render("- __*emphasise this*__")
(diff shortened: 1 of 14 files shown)


chrisjsewell:
I'm keeping commonmark as the default, but have added the option override and some other presets

markdown_it/tree.py
```diff
@@ -44,6 +45,12 @@ def __init__(self) -> None:
         # children (i.e. inline or img)
         self.children: List["SyntaxTreeNode"] = []
 
+    def __repr__(self) -> str:
+        return f"{self.__class__.__name__}({self.type})"
+
+    def __getitem__(self, item: int) -> "SyntaxTreeNode":
```
hukkinj1, line 51: I personally favor having just one way to do things, and would prefer the more explicit `.children[i]` syntax. Don't feel too strongly about that though.

If we keep this though, we might want to improve type annotations. `item` can be a `slice` in addition to `int` and return type will change accordingly too. An `@overload` is probably needed.

The return type will also break if subclassed. Merging https://github.com/executablebooks/markdown-it-py/pull/131 and changing the type to `_T` (and annotating `self` as `_T`) should fix this.

markdown_it/tree.py
```diff
@@ -44,6 +45,12 @@ def __init__(self) -> None:
         # children (i.e. inline or img)
         self.children: List["SyntaxTreeNode"] = []
 
+    def __repr__(self) -> str:
+        return f"{self.__class__.__name__}({self.type})"
```
hukkinj1, line 49: Dumb nitpick, but I would use `type(self)` for consistency since we already use that elsewhere in this class (also same reasons we use `str(self)` instead of `self.__str__()` etc.)
```suggestion
        return f"{type(self).__name__}({self.type})"
```

markdown_it/tree.py
```diff
@@ -76,6 +83,11 @@ def recursive_collect_tokens(
         recursive_collect_tokens(self, tokens)
         return tokens
 
+    @property
+    def is_root(self) -> bool:
```
hukkinj1, line 87: We could use `self.is_root` inside `def type()` definition to remove duplication.

markdown_it/tree.py
```diff
@@ -183,6 +195,23 @@ def _set_children_from_tokens(self, tokens: Sequence[Token]) -> None:
             )
             child._set_children_from_tokens(nested_tokens[1:-1])
 
+    def pretty(
```
hukkinj1, line 198: Should we print closing tags or add the slash character to make the tags self-closing if there's no content? That would make the output more or less valid XML and wouldn't confuse highlighters, parsers etc.

markdown_it/port.yaml
```diff
@@ -23,6 +23,11 @@
       In markdown_it/rules_block/reference.py,
       record line range in state.env["references"] and add state.env["duplicate_refs"]
       This is to allow renderers to report on issues regarding references
+    - |
+      The `MarkdownIt.__init__` signature is slightly different for updating options,
+      since you must always specify the config first, e.g.
+      use `MarkdownIt("commonmark", {"html": False})` instead of `MarkdownIt({"html": False})`
+    - The default configuration preset for `MarkdownIt` if "commonmark" not "default"
```
hukkinj1, line 30:
```suggestion
    - The default configuration preset for `MarkdownIt` is "commonmark" not "default"
```

docs/using.md
```diff
@@ -48,8 +49,15 @@ for token in md.parse("some *text*"):
...
-You can define this configuration *via* a preset name (`'zero'`, `'commonmark'` or `'default'`),
-or by directly supplying a dictionary.
+You can define this configuration *via* directly supplying a dictionary or a preset name:
+
+- `zero`: This configures the minimum components to parse text (i.e. just paragraphs and text)
+- `commonmark` (default): This configures the parser to strictly comply with the [CommonMark specification](http://spec.commonmark.org/).
+- `js-default`: This is the default in the JavaScript version.
+  Compared to `commonmark`, it disables HTML parsing and enables the table and strikethrough components.
+- `gfm-like`: This configures the parser to approximately comply with the [GitHub Flavored Markdown specification](https://github.github.com/gfm/).
+  Compared to `commonmark`, it disables HTML parsing and enables the table, strikethrough and linkify components.
```
hukkinj1, line 59: Should we add `mdit_py_plugins.tasklists` to this preset? That would make it pretty much GFM spec complete I think.

docs/using.md
```diff
@@ -97,6 +113,50 @@ Additionally `renderInline` runs the parser with all block syntax rules disabled
...
+```{code-cell}
+md = MarkdownIt("commonmark", {"typographer": True})
+md.enable(["replacements", "smartquotes"])
+md.render("'single quotes' (c)")
+```
+
+### Linkify
+
+The `linkify` component requires that [linkify-it-py](https://github.com/tsutsu3/linkify-it-py) be installed (e.g. *via* `pip install markdown-it-py[linkify]`).
+This allows URL links to be identified, without the need for enclosing in `<>` brackets:
```
hukkinj1, line 152:
```suggestion
This allows URI autolinks to be identified, without the need for enclosing in `<>` brackets:
```

I would explicitly mention autolink here, as normal links work just fine without the brackets even without linkify.

docs/using.md
```diff
@@ -183,22 +242,45 @@ This dictionary can also be deserialized:
...
+# Header
+
+Here's some text and an image ![title](image.png)
+
+1. a **list**
+
+> a *quote*
+""")
+
+node = SyntaxTreeNode.from_tokens(tokens)
```
hukkinj1, line 268: Just a heads-up, I made a PR (https://github.com/executablebooks/markdown-it-py/pull/132) which would change this syntax to `node = SyntaxTreeNode(tokens)` so if we merge that then one of the PRs needs to adapt depending on merge order.

markdown_it/main.py
```diff
@@ -89,21 +106,24 @@ def configure(self, presets: Union[str, Mapping]) -> "MarkdownIt":
...
-                )
-        if not presets:
-            raise ValueError("Wrong `markdown-it` preset, can't be empty")
-        config = AttrDict(presets)
-
-        if "options" in config:
-            self.set(config.options)
+            config = _PRESETS.get(presets, None)
+            if not config:
+                raise KeyError(f"Wrong `markdown-it` preset '{presets}', check name")
```
hukkinj1, line 111: Style suggestion:
```suggestion
            if presets not in _PRESETS:
                raise KeyError(f"Wrong `markdown-it` preset '{presets}', check name")
            config = _PRESETS[presets]
```

markdown_it/main.py
```diff
@@ -89,21 +106,24 @@ def configure(self, presets: Union[str, Mapping]) -> "MarkdownIt":
...
+        else:
+            config = presets
+
+        if not config:
+            raise ValueError("Wrong `markdown-it` config, can't be empty")
+
+        options = config.get("options", {}) or {}
+        if options_update:
+            options = {**options, **options_update}
```
hukkinj1, line 120: I'm fairly sure the JS upstream does a deep/recursive merge here. Might want to port that code?

E.g merging `{"a": {"b": 0, "c": 1}}` with `{"a": {"b": 99}}` will result in `{"a": {"b": 99}}` but I think `{"a": {"b": 99, "c": 1}}` is expected

markdown_it/main.py
```diff
@@ -27,36 +27,51 @@
...
-    }
-)
+_PRESETS = {
+    "default": presets.default.make(),
+    "zero": presets.zero.make(),
+    "commonmark": presets.commonmark.make(),
+}
+_PRESETS["js-default"] = _PRESETS["default"]
+_PRESETS["gfm-like"] = presets.default.make()
+_PRESETS["gfm-like"]["options"]["linkify"] = True
```
hukkinj1, line 37: Should we make these new presets public in `markdown_it.presets` package like the original ones are? That enables a workflow where you fetch a preset from there, mutate it, and pass it on to `.configure()`

hukkinj1 (changes requested):
Nice :+1: , here's my notes!

docs/using.md
hukkinj1: I'd almost go as far as calling the preset just `gfm` instead of `gfm-like` then. What do you think?

markdown_it/main.py
hukkinj1: Ah sorry it seems I misremember, upstream only does a "1st level merge" just like we do here.

markdown_it/main.py
```diff
@@ -27,36 +27,51 @@
...
 
 
 class MarkdownIt:
     def __init__(
-        self, config: Union[str, Mapping] = "commonmark", renderer_cls=RendererHTML
+        self,
+        config: Union[str, Mapping] = "commonmark",
+        options_update: Optional[Mapping] = None,
```
hukkinj1, line 44: One neat trick I've recently started to use in my APIs is
```python
from types import MappingProxyType

# This is basically just an immutable empty mapping
EMPTY_MAP = MappingProxyType({})

def func(options: Mapping = EMPTY_MAP):
    ...
```

This convention saves me a bunch of `Optional` annotations and `None` handling, and avoids the dangerous default problem.

Could be applied here if you like it :smile: You basically just need to create one global `EMPTY_MAP` instance per project and can reuse it.

docs/using.md
chrisjsewell: > Should we add mdit_py_plugins.tasklists to this preset?

No; because (a) the presets do not allow you to load extensions, and anyway (b) the point of moving the plugins to mdit_py_plugins is that markdown-it-py will no longer be dependent on them, in v1.0 mdit_py_plugins will be removed from the `install_requires`.
You could of course add a convenience function(s) in mdit_py_plugins to return a parser setup for full GFM compliance 😄 

> I'd maybe go as far as calling the preset just gfm instead of gfm-like then. What do you think?

No; because then I bet we will have issues opened that "the `gfm` preset behaves differently to when I use Markdown in GitHub". In fact, even with task lists, we are still missing one extension, which is the `tagfilter` extension for raw HTML: https://github.github.com/gfm/#disallowed-raw-html-extension- (currently HTML is just turned off)

On a side note, the good thing with CommonMark, and why I like it as the default, is that there is a comprehensive test suite (https://spec.commonmark.org/) that we test against (tests/test_cmark_spec/commonmark.json) and so can "guarantee" compliance. GFM says it builds in this spec, but the examples given for the extensions in https://github.github.com/gfm are not particularly extensive.

hukkinj1: I think I'd also enable HTML here, since most HTML is allowed in GFM, just a small subset of tags is disallowed. In fact, "commonmark" could be a better base preset for "gfm-like", as it'd be easy to compare/verify the changes against the annotated parts of the GFM spec.

docs/using.md
chrisjsewell: I'll have a look

markdown_it/main.py
chrisjsewell: thanks for the tip, I will have a look but probably not add it in this PR

markdown_it/tree.py
chrisjsewell: yeh exactly, I was going to leave this type annotation stuff to the other PR

markdown_it/tree.py
chrisjsewell: I'm not adding closing tags as the indentation denotes the nesting and this would make it overly verbose and less readable (which is the main point of this function). I could add self-closing, but TBH the highlighting in tests/test_tree/test_pretty.xml seems to be fine

docs/using.md
chrisjsewell: done in bc3a2efae6660a1a27ff28b540d6a7ee2bd7c430

markdown_it/main.py
chrisjsewell: done in bc3a2efae6660a1a27ff28b540d6a7ee2bd7c430

Issue #100, opened 2020-12-18
The way pypy version is specified in action/setup-py has changed.


tsutsu3:
https://github.com/actions/setup-python#specifying-a-pypy-version

https://github.com/executablebooks/markdown-it-py/pull/99#issuecomment-748159549

It might be better to change from `pypy3` to `pypy-3.6` or `pypy-3.7` (or both).

PR #138, opened 2021-03-04
🔧 MAINTAIN: Pin PyPy Python version

hukkinj1:
Closes https://github.com/executablebooks/markdown-it-py/issues/100

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -29,7 +29,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [pypy3, 3.6, 3.7, 3.8, 3.9]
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -29,7 +29,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [pypy3, 3.6, 3.7, 3.8, 3.9]
+        python-version: [pypy-3.6, 3.6, 3.7, 3.8, 3.9]
```
chrisjsewell, line 32: @hukkinj1 should this maybe be pypy-3.7 or 3.8, since 3.6 will be dropped at some point?

chrisjsewell (changes requested):

PR #139, opened 2021-03-05
🔧 MAINTAIN: Switch to declarative metadata

hukkinj1:
I was testing out some tooling ([setup-py-upgrade](https://github.com/asottile/setup-py-upgrade) and [setup-cfg-fmt](https://github.com/asottile/setup-cfg-fmt)) and came up with this auto-generated PR.

Declarative metadata has some advantages:
- Can be read by tooling without the need to execute arbitrary Python code
- Is easy for tools like [setup-cfg-fmt](https://github.com/asottile/setup-cfg-fmt) to modify. Brings us automatic version classifiers, ordering, etc.
- No need for some boilerplate like the `get_version` function.

Maybe we want to keep metadata in `setup.py` for consistency with other projects though?

I kept a dummy `setup.py` as pip doesn't yet support editable installs for projects without one.


chrisjsewell:
rather than installing setuptools, should it be defined in a pyproject.toml, e.g. https://github.com/aiidateam/aiida-core/blob/623d0d8b8b3f11e4f25a0d7d60061dccdba7150a/pyproject.toml#L1

this is the thing I don't like about this "new" setuptools system; you have to replace the single `setup.py` with three files to specify something I think should all be in `pyproject.toml` (https://discuss.python.org/t/where-to-get-started-with-pyproject-toml/4906/6)

chrisjsewell (changes requested):
see comment

hukkinj1:
> rather than installing setuptools, should it be defined in a pyproject.toml, e.g. https://github.com/aiidateam/aiida-core/blob/623d0d8b8b3f11e4f25a0d7d60061dccdba7150a/pyproject.toml#L1

So firstly, setuptools was already installed. The change made here merely updates it similar to what we do with `pip` because the version that was installed was affected by [this](https://github.com/pypa/setuptools/issues/1724) setuptools issue which made the build fail when using the `setup.cfg` mechanism for reading `__version__` from Python source. An up-to-date setuptools has this issue fixed.

Adding a pyproject.toml (thus activating PEP 517 build) and restricting setuptools version there would work as well yeah. This comes with the downside that we get yet another file as you mentioned :facepalm: :smile: 

> this is the thing I don't like about this "new" setuptools system; you have to replace the single setup.py with three files to specify something I think should all be in pyproject.toml (https://discuss.python.org/t/where-to-get-started-with-pyproject-toml/4906/6)

Yeah its pretty annoying right now. With [PEP 621](https://www.python.org/dev/peps/pep-0621/) accepted I guess setuptools will soon support metadata in pyproject.toml but we're not there yet. Meanwhile we have a few options:
- dont change anything
- merge this PR (declarative metadata)
  - with or without pyproject.toml (i.e. with or without PEP517 builds)
  - with or without the empty setup.py (i.e. with or without `pip install -e .` support)
- dump setuptools and use another PEP517/pyproject.toml native build tool (flit, poetry)

chrisjsewell:
Oh definitely sticking with setuptools, for editable install support, and because it is more well known for other contributors to work with.
Bit on the fence about the change, but if it is done at some stage I would certainly add the pyproject.toml build dependencies

hukkinj1:
> Oh definitely sticking with setuptools

:+1: 

> for editable install support, and because it is more well known for other contributors to work with.

The other two tools have editable installs, so its not that bad, just not via `pip`.

> Bit on the fence about the change, but if it is done at some stage I would certainly add the pyproject.toml build dependencies

Added the pyproject.toml

setup.py
```diff
@@ -1,73 +1,3 @@
-# from importlib import import_module
```
chrisjsewell, line 1: add a comment that this is required only for editable builds

chrisjsewell (changes requested):
Both mypy and flake8 configuration can be moved to `setup.cfg`, which might be desirable to reduce the number of files:

- https://flake8.pycqa.org/en/latest/user/configuration.html
- https://mypy.readthedocs.io/en/stable/config_file.html

hukkinj1:
Moved mypy and flake8 conf to setup.cfg

PR #142, opened 2021-03-11
🐛 FIX: Linkify link nesting levels

hukkinj1:
It seems to me like linkify link tokens have incorrect nesting levels. This PR fixes that.

~~This seems like a bug inherited from upstream. I'll make a quick PR there as well.~~ EDIT: On closer inspection this seems like a porting issue, so not making an upstream PR.


chrisjsewell:
thanks @hukkinj1
Hey @tsutsu3, since you wrote this code, perhaps you want to comment if you see any reason why this should not be case?

tsutsu3:
As @hukkinj1 pointed out, it is a porting error.
There is no problem with this fix.

chrisjsewell:
cheers!

PR #147, opened 2021-03-20
🧪 TESTS: Use pytest.xfail not pytest.skip

hukkinj1:
context: https://docs.pytest.org/en/documentation-restructure/how-to/skipping.html

I think `xfail` is more correct than `skip` for these tests.


chrisjsewell (approved):
Fair, although I'd rather they just didn't fail at all lol

PR #152, opened 2021-03-31
🚀 RELEASE: v1.0.0b1

chrisjsewell:


chrisjsewell:
its expected the documentation will fail for now, since myst-parser depends in markdown-it-py~=0.6.0

PR #153, opened 2021-03-31
Make `markdown_it.common.entities.entities` a dict. Remove AttrDict

hukkin:
Hey @chrisjsewell what's your thoughts on a change like this?

One test fails because of `AttDict` usage in `mdit-py-plugins`, so would need to be fixed there first. 

`mdit-py-plugins` uses `AttrDict` only this one time, and doesn't use `entities` at all. Not sure how much or if this would break much in MyST or elsewhere though? EDIT: MyST seems to not use `entities` at all, and the only use of `AttrDict` is `env` related, so should be removed regardless with `markdown_it_py>=1.0.0` support.


hukkinj1:
Made a [PR](https://github.com/executablebooks/mdit-py-plugins/pull/15) to remove AttrDict usage from mdit-py-plugins, so that would fix the test failure here.

hukkinj1:
On another note, maybe `entities` should be defined in `common/__init__.py` rather than `common/entities.py` ? The import path would more closely resemble JS upstream then, i.e. current `from markdown_it.common.entities import entities` vs `from markdown_it.common import entities`

chrisjsewell:
superseded by #181

PR #154, opened 2021-04-01
🔧 MAINTAIN: Update development packages

chrisjsewell:
The name of the package is already set up to use `markdown-it-py` in numerous places, but setup_cfg_fmt is hard-coded to convert these dashes to underscores https://github.com/asottile/setup-cfg-fmt/blob/de922db829605a067c7a407eb42620c3cb4466a0/setup_cfg_fmt.py#L365 🤷

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -44,8 +44,3 @@ repos:
     hooks:
     - id: mypy
       additional_dependencies: [attrs]
-
-  - repo: https://github.com/asottile/setup-cfg-fmt
-    rev: v1.16.0
-    hooks:
-    - id: setup-cfg-fmt
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -16,7 +16,7 @@ exclude: >
 repos:
 
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v3.3.0
+    rev: v3.4.0
     hooks:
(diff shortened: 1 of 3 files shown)


hukkinj1:
The underscored name will be normalized as per https://www.python.org/dev/peps/pep-0503/#normalized-names so it shouldn't be an issue. `setup-cfg-fmt` itself can be used to demonstrate: it [defines](https://github.com/asottile/setup-cfg-fmt/blob/de922db829605a067c7a407eb42620c3cb4466a0/setup.cfg#L2) underscored name in `setup.cfg` but the name shows hyphened on [PyPI page](https://pypi.org/project/setup-cfg-fmt/) and the package is installable with `pip install setup-cfg-fmt`.

[Here's](https://github.com/asottile/setup-cfg-fmt#normalizes-dashes-to-underscores-in-project-name) the reasoning why setup-cfg-fmt chooses to use underscores.

So personally I'd prefer keep the formatter (but won't be upset if you remove it :smile:).

chrisjsewell:
There's other things like the Conda build though, and yeh I can't be bothered to deal with potential side effects for the sake of a formatter.
If you could just turn this one change off, then I'd certainly want to keep it

hukkinj1:
In fact we could change the name to `markdown--_---......it-_py` lol. You can try `pip install markdown--_---......it-_py`, works perfectly :smile:

PR #160, opened 2021-04-28
👌 IMPROVE: Type annotate `common/utils`

hukkinj1:


markdown_it/common/utils.py
```diff
@@ -2,13 +2,14 @@
...
 import re
+from typing import Any
 
 from .entities import entities
 
 # from .normalize_url import unescape_string
 
 
-def charCodeAt(src: str, pos: int):
+def charCodeAt(src: str, pos: int) -> Any:
```
hukkinj1, line 12: The `Any` return type is to explicitly keep disabling type checking for the return value. This because the `None` case is pretty painful and dynamic: it happens when `int>=len(src)` but in code, before calling the function, we always have dynamic checks to ensure this doesn't happen. Thus the proper `Optional[int]` annotation requires many ugly `cast(int, ...)`s.

Best would be to raise IndexError instead of returning None, but I guess that isn't true to JS upstream, so IMO this is the second best option.

chrisjsewell: yeh I'm not picking at that thread right now 😬

chrisjsewell (approved):

PR #165, opened 2021-05-07
UPGRADE: attrs -> v21

tadeu:
Removing restriction for `attrs <21`. Only major breaking change is that `evolve` changed to not be recursive, but `evolve` is not used.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -29,7 +29,7 @@ project_urls =
 [options]
 packages = find:
 install_requires =
-    attrs>=19,<21
+    attrs>=19
     typing_extensions>=3.7.4;python_version<'3.8'
 python_requires = ~=3.6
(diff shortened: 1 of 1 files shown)


setup.cfg
```diff
@@ -29,7 +29,7 @@ project_urls =
 [options]
 packages = find:
 install_requires =
-    attrs>=19,<21
+    attrs>=19
```
chrisjsewell, line 32:
```suggestion
    attrs>=19,<22
```

cheers! I still want to pin this for future releases though; to guard against any nasty surprises later on

chrisjsewell (changes requested):

chrisjsewell:
For reference https://github.com/python-attrs/attrs/blob/main/CHANGELOG.rst

PR #168, opened 2021-07-07
⬆️ Update: Sync with markdown-it v12.1.0 and CommonMark v0.30

hukkin:
A port of this diff https://github.com/markdown-it/markdown-it/compare/df4607f1d4d4be7fdc32e71c04109aea8cc373fa...13cdeb95abccc78a5ce17acf9f6e8cf5b9ce713b

PR #169, opened 2021-07-12
🐛 FIX: Always suffix indented code block with newline

hukkin:
I'll propose this change upstream and add a link here soon. Probably don't want to merge before accepted upstream.

EDIT: Here's the upstream PR https://github.com/markdown-it/markdown-it/pull/799

### DIFF
--- a/markdown_it/port.yaml
+++ b/markdown_it/port.yaml
@@ -1,7 +1,7 @@
 - package: markdown-it/markdown-it
   version: 12.1.0
-  commit: 13cdeb95abccc78a5ce17acf9f6e8cf5b9ce713b
-  date: Jul 1, 2021
+  commit: e5986bb7cca20ac95dc81e4741c08949bf01bb77
+  date: Jul 15, 2021
   notes:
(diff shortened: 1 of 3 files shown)


hukkin:
Upstream PR is merged so this should be good to go. Would probably be good to do https://github.com/executablebooks/markdown-it-py/pull/168 first though, and then rebase this on master and update `port.yaml`.

chrisjsewell:
cheers, will try to have a look in the not too distant future

chrisjsewell:
I guess this can now be incorporated into a https://github.com/markdown-it/markdown-it/compare/12.1.0...12.2.0 sync

hukkin:
> I guess this can now be incorporated into a markdown-it/markdown-it@12.1.0...12.2.0 sync

Yeah. I don't have the time to do that right now, but did rebase and update port.yaml. So this should also be mergeable as is.

chrisjsewell:
>  I don't have the time to do that right now

tssk, so lazy 😜 yeh no worries, there is nothing in there that is a big priority and plenty busy myself

PR #171, opened 2021-07-20
♻️ REFACTOR: Port `mdurl` and `punycode` for URL normalisation

hukkin:
Closes https://github.com/executablebooks/markdown-it-py/issues/8


chrisjsewell:
Since https://github.com/hukkin/mdurl would become a core dependency of markdown-it, would it be ok to move this into executablebooks? and you would still be the primary administrator

hukkin:
Yeah on the same terms as mdformat I don't see any problem with that.

Note that there's a few TODOs still in mdurl:
- README.md needs an update
- Two test files need to be ported from JS to Python

Any help would be much appreciated!

PR #176, opened 2021-10-26
Explicity reexport MarkdownIt

paw-lu:
Explicitly export MarkdownIt.

Adds MarkdownIt to `__all__`.

This:

1. Allows markdown_it to pass strict mypy type checking, specifically `--no-implicit-reexport`
2. Removes the need for ignoring F401, as flake8 detects the use in `__all__`


chrisjsewell:
Now merged #120, but thanks for the PR!

PR #177, opened 2021-10-28
🔧 MAINTAIN: Use `pyproject-build` for package deployment

hukkin:
it's deprecated https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -119,8 +119,8 @@ jobs:
         python-version: 3.7
     - name: Build package
       run: |
-        pip install wheel
-        python setup.py sdist bdist_wheel
+        pip install build
+        python -m build
     - name: Publish
       uses: pypa/gh-action-pypi-publish@v1.1.0
       with:


chrisjsewell:
Thanks @hukkin, didn't know about this

Hey @pradyunsg, would you mind quickly confirming that this is now the "canonical" way to build packages? (before I roll it out to other packages)

hukkin:
I'm not pradyunsg :smile: , but the "canonical" replacements for direct setup.py invocations can be found in the Paul Ganssle blog post I linked (the table at the very bottom).

With PEP660 being a thing now (basically editable installs for PEP 517 packages) we could now also remove the setup.py since we only left it there for `pip install -e .` compatibility. We'd need to move to flit however, because setuptools [does not support the PEP yet](https://github.com/pypa/setuptools/issues/2816), flit does.

chrisjsewell:
> I'm not pradyunsg 😄 , but the "canonical" replacements for direct setup.py invocations can be found in the Paul Ganssle blog post I linked (the table at the very bottom).

Oh its not that I don't trust you lol, but pradsyung is a maintainer of pip, so it's nice to "hear it from the horse's mouth", before I start changing this across all my many repos 😅

> We'd need to move to flit however

yeh no, I tried this in some other repos (like https://github.com/aiidateam/archive-path), but there is a really annoying development issue for me; you have to use tox `isolated_build = True` (https://tox.wiki/en/latest/config.html#conf-isolated_build), which then means when "navigating" with e.g. VS Code pylance, it redirects to the built code rather than the source code (if that makes sense). It drives me crazy 😬

hukkin:
> but there is a really annoying development issue for me; you have to use tox isolated_build = True

Hmm, I suspect you may have issues in the future with further deprecations/removals related to legacy builds. And tox 4 will default to isolated builds. Maybe you want to configure tox to test source instead of built package (`skip_install` in tox). I don't think this is a recommended way, but may be the only way to have tooling work as you want in the future.

chrisjsewell:
> Hmm, I suspect you may have issues in the future with further deprecations/removals related to legacy builds. And tox 4 will default to isolated builds.

well I'll cross that bridge when I come to it lol; I'm sure I'm not the only one using a workflow including tox+vs code

pradyunsg:
Yea, using build instead of setup.py directly is a good idea. It also makes sure that your wheels are built from the sdist, which is a good thing.

chrisjsewell:
thanks, just wanted to double-check!

PR #184, opened 2022-01-16
🔧 MAINTAIN: Remove upper pin from `attrs`

hukkin:
`attrs` follows CalVer not SemVer, so upper pinning the "major" version shouldn't help much. It's year 2022 now so the next `attrs` version will be incompatible (With the upper pin that is. The code will be compatible because we see no `DeprecationWarning`s.)

See also: https://www.attrs.org/en/stable/changelog.html#changelog

> Versions follow CalVer with a strict backwards-compatibility policy.

> Put simply, you shouldn’t ever be afraid to upgrade attrs if you’re only using its public APIs. Whenever there is a need to break compatibility, it is announced here in the changelog, and raises a DeprecationWarning for a year (if possible) before it’s finally really broken.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -30,7 +30,7 @@ project_urls =
 [options]
 packages = find:
 install_requires =
-    attrs>=19,<22
+    attrs>=19
     mdurl~=0.1
     typing_extensions>=3.7.4;python_version<'3.8'
 python_requires = ~=3.6


chrisjsewell:
Eurgh, stupid calendar versioning. I'm not really a fan of having no upper pinning; it effectively renders all old version of a package broken, when the dependency does (inevitably) introduce back-incompatible changes

hukkin:
Yep, but since CalVer is in use the upper pin is somewhat arbitrary and has no guarantees on protecting against breaking changes.

Application devs can always add an upper pin themselves, but they cannot remove one set by library devs, so this can be very annoying.

chrisjsewell:
I think we should now be able to replace attrs with dataclasses?

hukkin:
Yeah Python 3.6 is no longer blocking it. I think we do a few things with `attrs` that `dataclasses` doesn't support so it's not a completely trivial change. At least `attr.evolve` comes to mind. Also `__slots__` support is new in Python 3.10 so can't be used, but while it's nice to have, I don't think it's essential or should be a blocker for using `dataclasses`. Or what does @chrisjsewell think?

chrisjsewell:
Yeh I'd certainly like to look into it: obviously attrs is nothing specific to the port, and was only used in lieu of a builtin solution

chrisjsewell:
attrs is now removed 🎉

Issue #185, opened 2022-01-24
An md file, or notebook cell, ending in a blockquote with a blank line,  causes a crash


fperez:
### Describe the bug

Consider a notebook that contains the following cell:

```
> test
>
```

or a markdown file with the above. The key point is that the cell/markdown content must _end with a blockquote that has a blank line at the end, still part of the block quote_. The python short traceback is (I'll attach the long one):

```python traceback
Exception occurred:
  File "/Users/fperez/local/conda/lib/python3.9/site-packages/markdown_it/rules_block/blockquote.py", line 158, in blockquote
    if state.srcCharCode[pos] == 0x20:  # /* space */
IndexError: tuple index out of range
```

### Reproduce the bug

Running a build with a notebook with the above (or an md file) should suffice.

### List your environment

```
> jupyter-book --version
Jupyter Book      : 0.11.3
External ToC      : 0.2.3
MyST-Parser       : 0.13.7
MyST-NB           : 0.12.3
Sphinx Book Theme : 0.1.10
Jupyter-Cache     : 0.4.3
NbClient          : 0.5.10
```

fperez:
Full Sphinx error log:

[sphinx-err-06zb0e7h.log](https://github.com/executablebooks/markdown-it-py/files/7922070/sphinx-err-06zb0e7h.log)

fperez:
@facusapienza21 - note this - Lab 1 for S159 was causing this crash and it took me a while to track down the issue.  Since it's a bug in markdown-it upstream, for now we can work around it by adding an extra blank line in cells that have block quotes, but I wanted you to be aware so you don't burn time on it...

hukkin:
>  Since it's a bug in markdown-it upstream

To clarify, I don't think this is an issue in upstream Javascript markdown-it, but rather a porting issue in markdown-it-py.

I've fixed this in https://github.com/executablebooks/markdown-it-py/pull/186

chrisjsewell:
Heya, yeh thanks, it is indeed a porting issue that I thought we had fully fixed already 😳.
Basically, for `array[outofbounds_index]` JS returns `null` (thus no crash) but Python raises an `IndexError`

fperez:
Thanks folks for the super-quick response! And yes @hukkin - I meant markdown-it-py, sorry for being imprecise in the js/py distinction :) 

I greatly appreciate the quick fix, we'll roll it out into our working image once it's part of the next release.

chrisjsewell:
[...]

PR #186, opened 2022-01-24
🐛 FIX: Crash when file ends with empty blockquote line

hukkin:
Closes https://github.com/executablebooks/markdown-it-py/issues/185

### DIFF
--- a/markdown_it/rules_block/blockquote.py
+++ b/markdown_it/rules_block/blockquote.py
@@ -154,8 +154,13 @@ def blockquote(state: StateBlock, startLine: int, endLine: int, silent: bool):
             # set offset past spaces and ">"
             initial = offset = state.sCount[nextLine] + 1
 
+            try:
+                next_char: Optional[int] = state.srcCharCode[pos]
+            except IndexError:
+                next_char = None
+
             # skip one optional space after '>'
-            if state.srcCharCode[pos] == 0x20:  # /* space */
+            if next_char == 0x20:  # /* space */
                 # ' >   test '
                 #     ^ -- position start of line here:
                 pos += 1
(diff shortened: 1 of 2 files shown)


chrisjsewell:
thanks @hukkin.
Given what I mentioned in https://github.com/executablebooks/markdown-it-py/pull/186, I'm thinking maybe we should just change https://github.com/executablebooks/markdown-it-py/blob/465b54d79641a3acd1d4e2ff483f1ccf4a4556ff/markdown_it/ruler.py#L50

so that it sets a "defaulttuple" that return None on out-of-bounds indexes?

chrisjsewell:
PS, ignore the docs error; that's a circular dependency issue that will be fixed when I release a new myst-nb

hukkin:
> I'm thinking now, maybe we should just change ... to defaulttuple

Yeah I'd like that as well. I would suggest
- Merge this to not delay fixing the known issue
- Work on defaulttuple in a separate PR. We have workarounds similar to the one in this PR elsewhere in the codebase as well, so there should an attempt to remove those in the defaulttuple PR.

> PS, ignore the docs error; that's a circular dependency

Yep, thanks for the heads up.

chrisjsewell (approved):

hukkin:
The only issue I see with the defaulttuple approach is that if `self.srcCharCode` was intended as a performance hack, then moving from tuple index access to a function call may not be great for perf at all.

chrisjsewell:
Could you not just create a simple https://docs.python.org/3/library/collections.abc.html#collections.abc.Sequence, that has `__getitem__` as a try/except

hukkin:
Hmm perhaps but I wonder if you lose all C level optimization of the tuple datatype. Should probably just measure performance and see what's relevant and what's not.

chrisjsewell:
Sure ting 👍

PR #190, opened 2022-01-24
Implement JS-esque `StateBase.srcCharCodeAt`

hukkin:
Replaces use of `StateBase.srcCharCode` with the more JS-esque `StateBase.srcCharCodeAt` that can return `None`.

I didn't remove `StateBase.srcCharCode` for ease of migration because mdit-py-plugins use it. But did deprecate it.

Happy to hear what you think @chrisjsewell . Not sure if this is the solution or migration strategy you want, but it's something :smile: 

I haven't measured performance yet. We may want to do that if this is something we want to proceed with otherwise

EDIT: This PR is in response to discussion in https://github.com/executablebooks/markdown-it-py/pull/186

### DIFF
--- a/markdown_it/helpers/parse_link_label.py
+++ b/markdown_it/helpers/parse_link_label.py
@@ -18,7 +18,7 @@ def parseLinkLabel(state: StateInline, start: int, disableNested: bool = False) 
     level = 1
 
     while state.pos < state.posMax:
-        marker = state.srcCharCode[state.pos]
+        marker = state.srcCharCodeAt(state.pos)
         if marker == 0x5D:  # /* ] */)
             level -= 1
             if level == 0:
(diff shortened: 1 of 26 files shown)


hukkin:
Seeing that we this PR only removes 4 instances of `try: ... except IndexError:` handling and that almost always we do an `if pos < maximum:` check or similar to ensure `None` is not an option, I actually still think it may be a fine decision to reject this change, and simply be more strict about out-of-range indices than upstream is.

The downsides for rejecting are diverging slightly from upstream logic, and the chance that there's still some place in the code where `None` handling is required but we don't know.

firefly9966 (approved):

chrisjsewell:
Closing as unrebaseable: `srcCharCode`, which this builds on, has since been removed from master, and the branch conflicts across most of the parser. The underlying question (should out-of-range reads return `None` like JS `charCodeAt`, or keep raising?) is still worth deciding; #415/#416 were another instance of that class of bug, so let's track it as an issue rather than this branch.

---

PR #200, opened 2022-02-19
Move from `attrs` to `dataclasses`

hukkin:
### BREAKING CHANGES
- `Token.as_dict`  -- Moved to using a simple `dataclasses.asdict` call. This means that many keyword arguments are removed from the method. Output is also slightly different (see `.yml` test regression files in this PR).
- ~~`Token.copy` -- Moved from `attr.evolve` to a simple `copy.copy` call. The stdlib `copy.copy` call does quite literally what the docstring of `Token.copy` says which is "return a shallow copy of the instance". Does `attr.evolve` do something extra? Is there a chance someone relies on this extra behavior?~~ EDIT: Answering my own question: Without keyword arguments, `attr.evolve` seems doesn't seem to do anything that `copy.copy` doesn't. I don't think this is a breaking change.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -52,4 +52,3 @@ repos:
     rev: v0.812
     hooks:
     - id: mypy
-      additional_dependencies: [attrs]
--- a/markdown_it/_compat.py
+++ b/markdown_it/_compat.py
@@ -1,0 +1,10 @@
+from __future__ import annotations
+
+from collections.abc import Mapping
+import sys
+from typing import Any
+
+if sys.version_info >= (3, 10):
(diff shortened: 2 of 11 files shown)


chrisjsewell:
Why not just use https://docs.python.org/3/library/dataclasses.html#dataclasses.replace

hukkin:
Oh wasn't aware such API existed.

However, since we don't use the kwargs or replace anything, I think I'd prefer `copy.copy` for readability. What do you think?

chrisjsewell:
Heya @hukkin I went ahead and did this in #212, keeping outputs the same.

Didn't add the slots compat, because I was lazy. Feel free to add

PR #202, opened 2022-03-09
📚 DOCS: Update usage guide to use PyPI package name

thibaudcolas:
It took me quite a few takes to realise `markdown_it` is the name of the Python module – the name of the package is `markdown-it-py` on PyPI. Hope this helps!


chrisjsewell (approved):
cheers

Issue #204, opened 2022-03-21
Combination of blockquotes, list and newlines causes an IndexError


anttimaki:
### Describe the bug

This is a bit of an edge case I encountered while rendering a README file. I was able to remove irrelevant parts, until just this was left:

```markdown
> QUOTE\n
+ UNORDERED LIST ITEM\n
  > INDENTED QUOTE\n
\n\n
```

This causes an IndexError. Removing any one of the three rows with content causes the example to work as expected. Also removing on or more newlines from the end causes the example to work. It's just this combination that causes problems.

### Reproduce the bug

```python
from markdown_it import MarkdownIt
md = MarkdownIt("gfm-like")
md.render("> QUOTE\n+ UNORDERED LIST ITEM\n  > INDENTED QUOTE\n\n\n")
```

Error:

```python
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/venv/lib/python3.8/site-packages/markdown_it/main.py", line 275, in render
    return self.renderer.render(self.parse(src, env), self.options, env)
  File "/venv/lib/python3.8/site-packages/markdown_it/main.py", line 260, in parse
    self.core.process(state)
  File "/venv/lib/python3.8/site-packages/markdown_it/parser_core.py", line 33, in process
    rule(state)
  File "/venv/lib/python3.8/site-packages/markdown_it/rules_core/block.py", line 14, in block
    state.md.block.parse(
  File "/venv/lib/python3.8/site-packages/markdown_it/parser_block.py", line 107, in parse
    self.tokenize(state, state.line, state.lineMax)
  File "/venv/lib/python3.8/site-packages/markdown_it/parser_block.py", line 77, in tokenize
    if rule(state, line, endLine, False):
  File "/venv/lib/python3.8/site-packages/markdown_it/rules_block/list.py", line 272, in list_block
    prevEmptyEnd = (state.line - startLine) > 1 and state.isEmpty(state.line - 1)
  File "/venv/lib/python3.8/site-packages/markdown_it/rules_block/state_block.py", line 137, in isEmpty
    return (self.bMarks[line] + self.tShift[line]) >= self.eMarks[line]
IndexError: list index out of range
```

### List your environment

* Python 3.8.10
* markdown-it-py 1.1.0 (also occurs on 2.0.1)

mib112:
May be related to #205 

Matching test case for this error (expected result taken from javascript markdon-it):

```
[...]

PR #207, opened 2022-04-13
🐛 FIX: Combination of blockquotes, list and newlines causes IndexError

hukkin:
Closes https://github.com/executablebooks/markdown-it-py/issues/204


markdown_it/rules_block/blockquote.py
```diff
@@ -296,6 +296,4 @@ def blockquote(state: StateBlock, startLine: int, endLine: int, silent: bool):
 
     state.blkIndent = oldIndent
 
-    state.lineMax += 1
```
hukkin, line 299: This increment doesn't exist in JS upstream so removing it https://github.com/markdown-it/markdown-it/blob/750a954d586c526e92bc4305bde2552743304d13/lib/rules_block/blockquote.js#L282

chrisjsewell (approved):
LGTM cheers

PR #213, opened 2022-04-15
🚀 RELEASE: v2.1.0

chrisjsewell:
@hukkin do you have any objections, regarding releasing this as 2.1 rather than 3.0?
Fo 99% of use cases, I don't think there should be a noticeable change, moving from attrs to dataclasses (it certainly did not affect myst-parser)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,19 @@
 # Change Log
 
+## 2.1.0 - 2022-04-15
+
+This release is primarily to replace the `attrs` package dependency,
+with the built-in Python `dataclasses` package.
+
+This should not be a breaking change, for most use cases.
+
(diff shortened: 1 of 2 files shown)


hukkin (commented):
I don't object either version.

Personally I'd do 3.0 mainly because of `Remove deprecated NestedTokens and nest_tokens`. The deprecation is a nice warning, but if we strictly follow SemVer, I don't think it makes the removals any less breaking.

hukkin:
If you plan to merge https://github.com/executablebooks/markdown-it-py/pull/214 it may be a good idea to do so before the release. Merging it afterwards will introduce a breaking change, as `__slots__` objects are stricter than objects without it (can't assign attributes dynamically).

PR #218, opened 2022-05-09
⬆️ UPGRADE: Allow linkify-it-py v2

hukkin:
A port of linkify-it 4.0.0 ([changelog here](https://github.com/markdown-it/linkify-it/blob/master/CHANGELOG.md#400--2022-04-22)) which makes a couple fixes and adds a new method


chrisjsewell:
Heya, could you briefly outline what the changes in v2 are and how it might affect parsing?

chrisjsewell:
better late than never 😅

PR #228, opened 2022-12-13
⬆️ UPGRADE: Allow up-to-date pytest-benchmark

hukkin:
Old versions of pytest-benchmark rely on the `py` package but don't list it as a dependency.

This fixes the broken `continuous-integration / benchmark` check.


chrisjsewell:
Fixed 👍

PR #229, opened 2023-01-10
📚 DOCS: Update the example

redstoneleo:
It is a very bad idea to `.disable('image')` in the example , most newbies just copy the example to test, by  `.disable('image')` causes unnecessary problem for them. I suffered the problem. So let's make a more user-friendly example.


chrisjsewell:
Thanks!

PR #254, opened 2023-03-14
🧪 Fix fuzzing test failures

chrisjsewell:
From https://github.com/google/oss-fuzz/tree/master/projects/markdown-it-py,
fixes issues 55363 and 55367

### DIFF
--- a/tests/test_fuzzer.py
+++ b/tests/test_fuzzer.py
@@ -1,0 +1,24 @@
+"""
+These tests are in response to reports from:
+https://github.com/google/oss-fuzz/tree/master/projects/markdown-it-py
+
+In the future, perhaps atheris could be directly used here,
+but it was not directly apparent how to integrate it into pytest.
+"""
+import pytest
+
+from markdown_it import MarkdownIt
+
+TESTS = {
+    55363: ">```\n>",
+    55367: ">-\n>\n>",
+    # 55371: "[](so&#4»0;!"  TODO this did not fail
+    # 55401: "?c_" * 100_000  TODO this does not fail
+}
+
+
+@pytest.mark.parametrize("raw_input", TESTS.values(), ids=TESTS.keys())
+def test_fuzzing(raw_input):
+    md = MarkdownIt()
+    md.parse(raw_input)
+    print(md.render(raw_input))
--- a/tests/test_fuzzer.py
+++ b/tests/test_fuzzer.py
@@ -13,7 +13,7 @@ TESTS = {
     55363: ">```\n>",
     55367: ">-\n>\n>",
     # 55371: "[](so&#4»0;!"  TODO this did not fail
-    # 55401: "?c_" * 100_000  TODO this does not fail
+    # 55401: "?c_" * 100_000  TODO this did not fail
 }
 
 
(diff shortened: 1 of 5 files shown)


tests/test_fuzzer.py
```diff
@@ -0,0 +1,24 @@
...
+"""
+import pytest
+
+from markdown_it import MarkdownIt
+
+TESTS = {
+    55363: ">```\n>",
+    55367: ">-\n>\n>",
+    # 55371: "[](so&#4»0;!"  TODO this did not fail
+    # 55401: "?c_" * 100_000  TODO this does not fail
```
chrisjsewell, line 16:
```suggestion
    # 55401: "?c_" * 100_000  TODO this did not fail
```

DavidKorczynski:
Thanks @chrisjsewell !

Would you be interested in moving the fuzzers from https://github.com/google/oss-fuzz/tree/master/projects/markdown-it-py upstream to here? Would be happy to create a set up that is easy for you to manage and also e.g. extend.

In the event you're curious, you can find a Fuzz Introspector report for `markdown-it-py` in this table: https://oss-fuzz-introspector.storage.googleapis.com/index.html where the current report (march 14th) links to: https://storage.googleapis.com/oss-fuzz-introspector/markdown-it-py/inspector-report/20230314/fuzz_report.html Here, there is yet another link to the "Project coverage" (the yellow button in top left corner) which shows code coverage achieved of the existing fuzzers: https://storage.googleapis.com/oss-fuzz-coverage/markdown-it-py/reports/20230314/linux/index.html -- 80% atm

chrisjsewell:
Hey @DavidKorczynski 👋🏼

> Would you be interested in moving the fuzzers from https://github.com/google/oss-fuzz/tree/master/projects/markdown-it-py upstream to here? Would be happy to create a set up that is easy for you to manage and also e.g. extend.

Yep that would be great thanks 🙏.
As it mentions in the docstring, I had a look, but didn't see any easy way to integrate it with pytest.
How long does the fuzzing take to run? I tried locally but it seemed to be running for a very long time. Obviously I don't want to have to wait ages every time I want to check the tests. In that case I would have them separate to the current unit tests.

DavidKorczynski:
The fuzzers are meant to be run for a long period of time since the longer they run the more code they explore. There is currently a start-up overhead due to the use of `atheris.instrument_all` which can take a while (~45 sec?)

There are probably a few different ways to intergrate -- what do you think of the following?
```python
import sys
import atheris

@pytest.mark.parametrize(
    "data",
    [
        b"\x42\65\xff.... (fuzzertest 1 case, check for regressions)",
        b"fuzzertest 2 case, check for regressions"
    ]
)
def TestOneInput(data):
  fdp = atheris.FuzzedDataProvider(data)
  md = MarkdownIt()
  raw_markdown = fdp.ConsumeUnicodeNoSurrogates(sys.maxsize)
  md.parse(raw_markdown)
  md.render(raw_markdown)


def main():
  atheris.instrument_all()
  atheris.Setup(sys.argv, TestOneInput)
  atheris.Fuzz()


if __name__ == "__main__":
  main()
```

Using the above solution you can just paste the byte arrays in from the OSS-Fuzz reports as `pytest` inputs to `TestOneInput`, which can be used as regression testing when run under pytest. OSS-Fuzz will continue to run `main` for each fuzzer and thus be able to pass arbitrary input to `TestOneInput`.

PR #255, opened 2023-03-15
🧪 Add OSS-Fuzz set up

DavidKorczynski:
This is a follow-up to https://github.com/executablebooks/markdown-it-py/pull/254#issuecomment-1468456515

@chrisjsewell I added some documentation which may be helpful to navigate the OSS-Fuzz infra. I hope this should make it easier to e.g. experiment and test with the fuzzing.

I'm the original author of the fuzzers and am happy to license them according to the MIT of this repository.


chrisjsewell:
Thanks @DavidKorczynski!
As you can see, I just added a workflow to get an idea of how it runs; obviously it looks like it takes too long to run for all PRs, and would so would just be run as a maybe weekly cron job

From the `README.md`, it is not quite clear what the difference is between running the files in `tests/fuzz` directly, e.g. `python tests/fuzz/fuzz_markdown.py`, and following the steps to run using the `git clone https://github.com/google/oss-fuzz` steps, e.g. `python infra/helper.py run_fuzzer markdown-it-py fuzz_markdown`
What is the difference?
Is the code in `google/oss-fuzz` more thorough?

For the actual code, I'd like to understand it a little more, the language support is a little lacking 😅:

image

- Is there a simple way to explain what "instrumenting" actually does?
- what can you actually pass with `sys.argv`, I tried doing things like `--help` and it didn't seem to actually do anything
- can you restrict the fuzzing to only a certain module or something? or does it always have to run through everything?
  - is the way I run it in the workflow even right, or is it running through the whole of python's core code or something 😅

chrisjsewell:
so yeh, the fuzzing hit the 6 hours time limit, is that right!?

DavidKorczynski:
> so yeh, the fuzzing hit the 6 hours time limit, is that right!?

When run as follows `python3 ./fuzz_markdown.py` the fuzzer will run forever or until a crash is found.

You can limit the fuzzer time using `-max_total_time=X` where `X` is the number of seconds to run. `python3 ./fuzz_markdown.py -max_total_time=10` runs the fuzzer for 10 seconds where the timer starts post instrumentation.

> what can you actually pass with sys.argv, I tried doing things like --help and it didn't seem to actually do anything

atheris uses libFuzzer under the hood, which means you can pass the commandline flags specified here: https://llvm.org/docs/LibFuzzer.html

> Is there a simple way to explain what "instrumenting" actually does?

The instrumentation adds code that essentially correspond to counters throughout the Python code. These counters will be used by the fuzzer to trace what code was executed by a given input, and further uses this to check whether a input executes unique parts of `markdown-it`. This is needed because the fuzzer will find unique inputs that trigger unique parts of the code. Fuzzing is a "genetic mutational algorithm" and the instrumentation is used to separate which which seeds to save, I talk a small bit about here https://youtu.be/zIyIZxAZLzo?t=368 and from [06:08 - 10:30] approximately.

> From the README.md, it is not quite clear what the difference is between running the files in tests/fuzz directly, e.g. python tests/fuzz/fuzz_markdown.py, and following the steps to run using the git clone https://github.com/google/oss-fuzz steps, e.g. python infra/helper.py run_fuzzer markdown-it-py fuzz_markdown
What is the difference?
Is the code in google/oss-fuzz more thorough?

No difference as such. It's the same program.

OSS-Fuzz will, however, wrap the code in a package using `pyinstaller` and offers some additional things e.g. code coverage visualisation to check how much of the target code is executed.

I would not necessarily recommend setting up a cron-job in the CI, as OSS-Fuzz will make sure the fuzzers run continuously. If anything, I'd advice to set it up in a way that tests if the fuzzers work (i.e. run for a short period of time) or use the CIFuzz job that comes accompanied with OSS-Fuzz: https://google.github.io/oss-fuzz/getting-started/continuous-integration/#integrating-into-your-repository

DavidKorczynski:
> I don't think that I've quite mastered this yet 😅 what did I do wrong?

I think it's the `language` attribute missing: https://google.github.io/oss-fuzz/getting-started/continuous-integration/#optional-configuration

If you set it to `python` I think it should work!

chrisjsewell:
> are crashes only failing/uploaded if they are caused specifically by the changed code in the PR?

oh yeh I should properly read the documentation lol: https://google.github.io/oss-fuzz/getting-started/continuous-integration/#how-it-works

DavidKorczynski:
> it seems there was a crash, but it did not upload it, so I can't reporoduce it?
> are crashes only failing/uploaded if they are caused specifically by the changed code in the PR?

Yes -- this given crash was visible before, it seems to be this one: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=55371

chrisjsewell:
ok great, so from your readme/explanation, it seems:

1. the workflow could be helpful, for "early detection" of failures in PRs, although perhaps I don't want to wait 10 minutes for every PR commit 🤔
2. The command: `python3 infra/helper.py reproduce markdown-it-py {FUZZER_NAME} {PATH_TO_MINIMIZED_TESTCASE}`, seems the most useful, for quickly reproducing bugs reported by you guys. Perhaps I can set up a "tox env" to git clone + run this

I'm not sure of the usefulness of `tests/fuzz/fuzz_markdown.py` and `tests/fuzz/fuzz_markdown_extended.py` though; I'm not sure how/if I would actually ever use them

DavidKorczynski:
1. Many projects prefer to have a shorter runtime -- I personally have mine on 120sec, but definitely tailor it to your needs and what fits your workflow. OSS-Fuzz will run the fuzzers continuously which will catch bugs that are more complex to reach.

> I'm not sure of the usefulness of tests/fuzz/fuzz_markdown.py and tests/fuzz/fuzz_markdown_extended.py though; I'm not sure how/if I would actually ever use them

You're not meant to use them as such. Having a folder with fuzzers in this repo that OSS-Fuzz then fetches to run continuously serves the purpose of you being able to add/modify the fuzzers without having to make PRs on the OSS-Fuzz repo.

chrisjsewell:
Hey @DavidKorczynski thanks for your patience 😅 

In 2d46a43bd7a55310452b33f9ed2801f7b5cc4dca and d1852a5f8e2cd33602d670cf602593e664c7a2a7 I separately added the CI workflow for PRs and a tox environment to reproduce crash files,

Then here, in 497fee82a110f39dca98a7009bffc5a3710f9a5b, I rewrote the README as I now understand the process.
If you could have a look and check that this makes sense, then I think this is good to merge!

PR #258, opened 2023-03-24
fix: specify release bound for compat test

agoose77:
I didn't bisect to find when 3.10 adds the `slots` argument; I know it's in the release itself, and I don't think we need to be *that* careful to provide features to old alpha/beta/rc releases.

### DIFF
--- a/markdown_it/_compat.py
+++ b/markdown_it/_compat.py
@@ -4,7 +4,7 @@ from collections.abc import Mapping
 import sys
 from typing import Any
 
-if sys.version_info >= (3, 10):
+if sys.version_info >= (3, 10, 0, "release"):
     DATACLASS_KWARGS: Mapping[str, Any] = {"slots": True}
 else:
     DATACLASS_KWARGS: Mapping[str, Any] = {}
--- a/markdown_it/_compat.py
+++ b/markdown_it/_compat.py
@@ -4,7 +4,7 @@ from collections.abc import Mapping
 import sys
 from typing import Any
 
-if sys.version_info >= (3, 10, 0, "release"):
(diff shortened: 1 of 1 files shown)


chrisjsewell:
Thanks @agoose77, @hukkin wrote this particular part of the code, and so I would suggest he should have a say on this change before merging

markdown_it/_compat.py
```diff
@@ -4,7 +4,7 @@
 import sys
 from typing import Any
 
-if sys.version_info >= (3, 10):
+if sys.version_info >= (3, 10, 0, "release"):
```
hukkin, line 7:
```suggestion
if sys.version_info >= (3, 10, 0, "final"):
```

It seems like `"release"` is not a valid release level. I think what you want is `"final"`? https://docs.python.org/3/library/sys.html#sys.version_info

hukkin (changes requested):
This should work, although my opinion is that support for alpha/beta releases should not be expected, so perhaps not worth it to further complicate the tuple comparison with non-integer values?

markdown_it/_compat.py
agoose77: :clown_face: :facepalm: 

Wow, yes. Great catch @hukkin!

I will cut myself some slack and say that I was fairly sleep deprived when I wrote this.

firefly9966 (approved):

chrisjsewell:
Closing as no longer applicable: the `_compat.py` version guard this patched has since been removed entirely, and `requires-python` is now `>=3.10`, so pre-release 3.10 builds are out of scope. Thanks!

---

PR #296, opened 2023-07-03
📚 DOCS: Clarify docs regarding security configuration

### LINKED ISSUE #295: Default configuration is unsafe, contrary to documentation
I'm using `markdown-it-py` on a new Django project and it's perfect, but I think the claims regarding security in the docs are quite misleading.

Both the [PyPI project description](https://pypi.org/project/markdown-it-py/) and [readthedocs home page](https://markdown-it-py.readthedocs.io/en/latest/) say that this parser-renderer is "safe by default", but it looks like those statements were just copied over verbatim from the original JS project. The [dedicated page on security](https://markdown-it-py.readthedocs.io/en/latest/security.html#md-security) further says HTML is disabled by default as a blanket protection, but on closer inspection that actually describes the original JavaScript `markdown-it` project, **not this Python port, which switches the default configuration to unsafe full CommonMark compliance** (meaning, HTML parsing is enabled, potentially allowing for injection-style attacks when used in web apps).

#56 is proof of this issue, as the reporter said "I wanted to go for markdown-it for security stance" but the code snippets show usage of the Python package without any deliberate configuration for safer parsing. To be fair, that is easy: just pass `"js-default"` to the constructor. But the security page in the docs does not mention this is required to actually get the benefits, and all the "safe by default" claims imply that no such effort is needed at all.

This is at least more of a documentation issue than an actual vulnerability in the code, and I can submit a PR for the necessary edits, but I'm not sure if I'm missing something. Why was the default config switched to full CommonMark compliance in the first place? The docs imply that this Python port aims to minimize differences with the JS version otherwise. Perhaps the motivations for the port mean that security is less of a priority? (That seems to be the case with Executable Books which appears to be concerned more with data science tooling than public-facing web applications, but I'm not too familiar with the organization.) If so, that is okay, but these priorities and design principles should be reflected accurately in the project pages and docs.

(I'm sorry in case my tone comes across as not nice; I'm actually grateful for this project! I just think the issue needs a bit more attention, and I believe in the importance of good documentation.)

djramones:
Fixes #295.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@
 - Configurable syntax: you can add new rules and even replace existing ones.
 - Pluggable: Adds syntax extensions to extend the parser (see the [plugin list][md-plugins]).
 - High speed (see our [benchmarking tests][md-performance])
-- [Safe by default][md-security]
+- Easy to configure for [security][md-security]
 - Member of [Google's Assured Open Source Software](https://cloud.google.com/assured-open-source-software/docs/supported-packages)
 
 This is a Python port of [markdown-it], and some of its associated plugins.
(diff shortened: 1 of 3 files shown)


chrisjsewell (commented):
Just a minor change; "does not care about security" felt a bit harsh 😅

chrisjsewell (approved):
cheers!

PR #310, opened 2023-11-21
Isort is already included in ruff, use that.

venthur:


chrisjsewell:
Heya, this won't actually sort them when running pre-commit though, it'll just fail if they are not sorted. You need to set up the ruff formatted and stuff for that

venthur:
Good call. I only use `--fix` here to fix the linter issues that can be automatically fixed. In case of `isort` all of them should be fixable, some others might not be, though, and we're running `black` anyways which should take care of the rest. In principal, you could think about replacing `black` with `ruff` as well and go full `ruff format .`

chrisjsewell:
For formatting, you shouldn't have to run anything apart from pre-commit, and it should just do it all

For linting obviously that's not always possible.

Sorting imports is certainly formatting and not linting though

venthur:
Ah ok, i wasn't aware. Feel free to close that PR then. Cheers!

chrisjsewell:
Superceded by #313 😄

PR #336, opened 2024-11-02
👌 Throw an error if 3rd party plugin doesn't increment `line` or `pos` counters

hukkin:
(previously, markdown-it would likely go into infinite loop instead)

This really is a port of upstream changes https://github.com/markdown-it/markdown-it/compare/13.0.1...49ca65bbef067c7dba63468a48c4aee3048607dc which adds up to just one changelog entry (the PR title).


markdown_it/common/utils.py
```diff
@@ -91,25 +91,24 @@ def fromCodePoint(c: int) -> str:
     r'\\([!"#$%&\'()*+,\-.\/:;<=>?@[\\\]^_`{|}~])' + "|" + r"&([a-z#][a-z0-9]{1,31});",
     re.IGNORECASE,
 )
-DIGITAL_ENTITY_BASE10_RE = re.compile(r"#([0-9]{1,8})")
-DIGITAL_ENTITY_BASE16_RE = re.compile(r"#x([a-f0-9]{1,8})", re.IGNORECASE)
+DIGITAL_ENTITY_TEST_RE = re.compile(
```
chrisjsewell, line 94: This change seems separate to the PR title?
Can you split it into a separate PR

hukkin:
It seems the commit message got somehow malformed when copy pasting a message with backticks to to terminal. Perhaps best to squash and merge with the PRs (or other) message.

chrisjsewell (commented):
Heya

markdown_it/common/utils.py
hukkin: Most changes here are unrelated refactors actually. This is a port of the diff in the first post. Should something else be split too?

chrisjsewell: Yeh ideally it should be split up fully into "logical changes" 😅

hukkin: :+1: Pushed a new commit history. Would this split work?

chrisjsewell: > Would this split work?

Please split them up into separate PRs, with a clear description of their intention; the smaller the PR better, and the more likely I can just merge them without much effort/delay

PR #338, opened 2024-11-14
👌 Add "store_labels" to OptionsType

hukkin:
Type checker will otherwise complain when accessing this key.


chrisjsewell:
done in #343

PR #382, opened 2026-02-03
🔧 Add typing to Scanner

Alunderin:
Changes `state_inline.Scanner` into a `typing.NamedTuple` instead of a `collections.namedtuple`. They should be functionally equivalent except that the former has typing for its members while the latter does not.

I tend to use strict type checking for my own projects so it'd be nice if Pyright didn't freak out when I try to access the result of `scanDelims` without using `typing.cast`.


chrisjsewell (approved):
cheers!

PR #387, opened 2026-04-06
Fix CJK emphasis delimiter detection in scanDelims

awoni:
## Summary
Fix emphasis delimiter detection for mixed CJK/ASCII text.

In CommonMark's `scanDelims`, when an emphasis closing marker `**` is preceded by ASCII punctuation (e.g. `%`) and followed by a CJK character, `right_flanking` evaluates to `False` — causing bold to silently fail.

Example: `湾岸の**46%**を` renders as plain text instead of `<strong>46%</strong>`.

## Root Cause
CJK characters (e.g. `を`, `の`, `は`) are not classified as whitespace or punctuation by the spec. When `lastChar='%'` (punctuation) and `nextChar='を'` (neither whitespace nor punctuation):

`right_flanking = not (isLastPunctChar and not (isNextWhiteSpace or isNextPunctChar))`

evaluates to `False`, so the closing `**` is rejected.

## Fix
Treat characters above U+2E7F as punctuation for flanking delimiter checks (6 lines added to `scanDelims` in `state_inline.py`). This covers CJK Unified Ideographs, Hiragana, Katakana, Hangul, and other East Asian scripts.

## Test Cases
```
md = MarkdownIt("commonmark")

assert "<strong>" in md.render("湾岸の**46%**を供給")
assert "<strong>" in md.render("の**20〜25%**が通過する")
assert "<strong>" in md.render("日本語の**太字**テスト")
assert "<strong>" in md.render("**全部太字**")
assert "<strong>" in md.render("English **bold** test")
```
All pass. Existing English-only behavior is unchanged.

## Note
The same bug likely exists in the JavaScript version of markdown-it.

### DIFF
--- a/markdown_it/rules_inline/state_inline.py
+++ b/markdown_it/rules_inline/state_inline.py
@@ -148,6 +148,14 @@ class StateInline(StateBase):
         isLastWhiteSpace = isWhiteSpace(ord(lastChar))
         isNextWhiteSpace = isWhiteSpace(ord(nextChar))
 
+        # Treat CJK ideographs as punctuation for flanking delimiter checks
+        # so that emphasis works correctly in mixed CJK/ASCII contexts.
+        # e.g. 湾岸の**46%**を should render <strong>46%</strong>
+        if not isNextPunctChar and ord(nextChar) > 0x2E7F:
+            isNextPunctChar = True
+        if not isLastPunctChar and ord(lastChar) > 0x2E7F:
+            isLastPunctChar = True
+
         left_flanking = not (
             isNextWhiteSpace
             or (isNextPunctChar and not (isLastWhiteSpace or isLastPunctChar))


chrisjsewell:
## PR #387 Review: Fix CJK emphasis delimiter detection in scanDelims

### What it does

The PR modifies `scanDelims` in state_inline.py to treat any character with codepoint > U+2E7F as "punctuation" for flanking delimiter checks. This fixes cases like `湾岸の**46%**を` where the closing `**` couldn't be recognized as right-flanking because `を` (a CJK character) is neither whitespace nor punctuation under the spec rules.

---

### Concerns

**1. Deviation from CommonMark spec (not yet adopted upstream)**

The CommonMark spec defines emphasis flanking using Unicode categories `P` (punctuation) and `S` (symbols). CJK letters are category `Lo` — regular characters. The spec issue [commonmark/commonmark-spec#650](https://github.com/commonmark/commonmark-spec/issues/650) has been open since 2020 with no resolution. The upstream JS markdown-it has a corresponding PR ([markdown-it/markdown-it#1145](https://github.com/markdown-it/markdown-it/pull/1145)) that is **not merged**, and a commenter (@tats-u, author of `markdown-it-cjk-friendly`) noted *"such a change must be merged into CommonMark first."*

Merging this makes markdown-it-py diverge from both the spec and its upstream JS reference implementation.

**2. Overly broad heuristic**

The check `ord(char) > 0x2E7F` treats **all** characters above codepoint 11903 as punctuation-like. While most blocks immediately above U+2E7F are CJK-related (CJK Radicals, Kangxi, Kana, Hangul, CJK Unified Ideographs), this also captures:
- Yi Syllables (U+A000–A48F)
- Vai (U+A500–A63F)
- Various African and South/Southeast Asian scripts at higher codepoints
- Emoji and miscellaneous symbols

By contrast, the JS markdown-it PR #1145 uses a targeted helper checking only CJK Unified Ideographs (U+4E00–9FFF) and Hangul Syllables (U+AC00–D7A3) — and was still asked to add Hiragana/Katakana.

**3. No tests committed**

**4. Semantic mismatch in the fix**

The comment says "Treat CJK ideographs as punctuation" but the effect is more subtle — by marking CJK as punctuation, the fix changes the flanking formula's boundary condition. The upstream JS approach is cleaner: it adds `|| isNextCJK` directly to the `right_flanking` formula rather than pretending CJK characters are punctuation (which also affects `can_open`/`can_close` via `isLastPunctChar`/`isNextPunctChar` in ways that may have unintended side effects).

---

### Comparison with other parsers

| Parser | Approach | Status |
|--------|----------|--------|
| **CommonMark spec** | CJK letters are regular characters, not punctuation | Spec issue #650 open, no change |
| **JS markdown-it** | Follows spec strictly | PR #1145 (relaxed right-flanking for CJK) open, **not merged** |
| **markdown-it-cjk-friendly** | Plugin with precise Unicode range checks | Active, recommended as opt-in |
| **Python-Markdown** | Regex-based; proposed `cjk_friendly_emphasis` extension | Rejected for core, suggested as third-party |
| **mistune** | Regex-based, asterisks more permissive | No explicit CJK handling |
| **This PR** | Broad `> 0x2E7F` heuristic in core | Deviates from all above |

---

### Recommendation

This PR addresses a real user-facing problem but the approach has issues:

1. **Should be a plugin, not a core change** — consistent with the ecosystem consensus (CommonMark, JS markdown-it, Python-Markdown all treat this as opt-in behavior)
2. **If merged into core**, the heuristic should be replaced with explicit Unicode range checks (Hiragana U+3040–309F, Katakana U+30A0–30FF, CJK Unified U+4E00–9FFF, Hangul U+AC00–D7A3, CJK Ext-A U+3400–4DBF at minimum)
3. **The fix logic should modify the flanking formula directly** (like the JS PR does with `|| isNextCJK`) rather than overloading the punctuation flag, which has broader side effects on `can_open`/`can_close`
4. **Tests are required** regardless of approach

tats-u:
This sloPR is far too naive for something created by Opus or Sonnet. You should rewrite it from scratch, providing appropriate and sufficient background information/context. In the first place, the original markdown-it is in a sort of maintenance mode, and this repository probably doesn't plan to incorporate any extensions beyond those used in Git**Hu**b-Flavored Markdown into core. (CJK-Friendly Emphasis is available in Git**La**b)

tats-u:
You can take test cases from https://github.com/tats-u/markdown-cjk-friendly and reference implementations from there, https://github.com/kivikakk/comrak/pull/582, https://github.com/xoofx/markdig/pull/921, and (rejected) https://github.com/yuin/goldmark/pull/529 if you can afford to burn input tokens.

chrisjsewell:
Closing, per the earlier review: this changes emphasis parsing for every codepoint above U+2E7F (which also captures many non-CJK scripts and emoji), diverges from both the CommonMark spec and upstream markdown-it, and has no tests. The ecosystem answer for this is an opt-in plugin such as markdown-it-cjk-friendly; if CommonMark or upstream adopt CJK-aware flanking rules we will port them. Thanks for raising it.

---

PR #388, opened 2026-05-03
✨ Add `gfm-like2` preset with task lists, alerts, and single-tilde strikethrough

chrisjsewell:
### Summary

Adds a new `gfm-like2` preset that extends `gfm-like` with three GFM features:

- **Task lists** — `- [x] done` / `- [ ] todo` checkbox syntax in list items
- **Alerts** — `> [!NOTE]`, `> [!TIP]`, `> [!WARNING]`, etc. inside blockquotes
- **Single-tilde strikethrough** — `~text~` in addition to `~~text~~`

These are enabled via the `gfm-like2` preset or individually through the `tasklists`, `alerts`, and `strikethrough_single_tilde` options. The existing `gfm-like` preset is unchanged, so as to remain back-compatible.

### Why in markdown-it-py, not mdit-py-plugins?

Task lists and alerts are implemented by integrating detection directly into the existing block-level parsers (list.py and blockquote.py), rather than as post-processing rules:

- Checkbox detection happens during list item parsing, before the sub-parser runs on the item content
- Alert detection happens during blockquote parsing, before the inner content is tokenized

This design is not achievable from a plugin: plugins can only add new rules or post-process the token stream — they cannot modify the internals of `list_block()` or `blockquote()` to inject detection at the right point in the parsing pipeline. Implementing these as post-processing core rules would work functionally, but it means re-walking and mutating the token stream after the fact, which is less clean and less consistent with how the block parsers are designed to work.

Single-tilde strikethrough similarly extends the existing strikethrough rule's matching logic (opener/closer width matching), which is more naturally done inside the rule than bolted on externally.

### Changes

- **`rules_block/list.py`** — Detect `[ ]`/`[x]`/`[X]` at content start during list item parsing; set `token.meta["checked"]`; advance `bMarks` past the checkbox; add CSS classes (`task-list-item`, `contains-task-list`) after the list loop
- **`rules_block/blockquote.py`** — Detect `[!TYPE]` on the first content line; emit `alert_open`/`alert_close` + title tokens instead of `blockquote_open`/`blockquote_close`; skip the marker line during tokenization
- **`rules_inline/strikethrough.py`** — When `strikethrough_single_tilde` is enabled, accept 1 or 2 tildes (reject 3+); enforce opener/closer width matching in `_postProcess`
- **renderer.py** — Add `list_item_open` render method that injects checkbox HTML when `meta["checked"]` is present
- **`presets/__init__.py`** — Add `gfm_like2` preset class
- **`main.py`** — Register `gfm-like2` in `_PRESETS`
- **utils.py** — Add `tasklists`, `alerts`, `strikethrough_single_tilde` keys to `OptionsType`
- **pyproject.toml** — Add `pytest-timeout` to test deps; set 10s default timeout
- **Test fixtures** — 11 tasklist cases, 15 alert cases, 13 single-tilde strikethrough cases

### Usage

```python

md = MarkdownIt("gfm-like2")
md.render("- [x] done\n- [ ] todo")
md.render("> [!NOTE]\n> This is a note.")
md.render("~strikethrough~")
```

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -6,6 +6,8 @@ The following plugins are embedded within the core package:
 
 - [tables](https://help.github.com/articles/organizing-information-with-tables/) (GFM)
 - [strikethrough](https://help.github.com/articles/basic-writing-and-formatting-syntax/#styling-text) (GFM)
+- [task lists](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/about-task-lists) (GFM) — `- [x] done`
+- [alerts](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts) (GFM) — `> [!NOTE]`
 
 These can be enabled individually:
 
@@ -18,7 +20,8 @@ or as part of a configuration:
 
 ```python
 from markdown_it import MarkdownIt
-md = MarkdownIt("gfm-like")
+md = MarkdownIt("gfm-like")   # tables, strikethrough, linkify
+md = MarkdownIt("gfm-like2")  # + task lists, alerts, single-tilde strikethrough
 ```
 
 ```{seealso}
--- a/docs/using.md
+++ b/docs/using.md
@@ -58,6 +58,9 @@ You can define this configuration *via* directly supplying a dictionary or a pre
 - `gfm-like`: This configures the parser to approximately comply with the [GitHub Flavored Markdown specification](https://github.github.com/gfm/).
   Compared to `commonmark`, it enables the table, strikethrough and linkify components.
   **Important**, to use this configuration you must have `linkify-it-py` installed.
+- `gfm-like2`: Builds on `gfm-like` and additionally enables task lists (`- [x] done`),
+  GitHub-style alerts (`> [!NOTE]`), and single-tilde strikethrough (`~text~`).
+  **Important**, to use this configuration you must have `linkify-it-py` installed.
 
 ```{jupyter-execute}
 from markdown_it.presets import zero
--- a/markdown_it/main.py
+++ b/markdown_it/main.py
@@ -26,6 +26,7 @@ _PRESETS: dict[str, PresetType] = {
     "zero": presets.zero.make(),
     "commonmark": presets.commonmark.make(),
     "gfm-like": presets.gfm_like.make(),
+    "gfm-like2": presets.gfm_like2.make(),
 }
 
 
--- a/markdown_it/presets/__init__.py
+++ b/markdown_it/presets/__init__.py
@@ -1,4 +1,4 @@
-__all__ = ("commonmark", "default", "gfm_like", "js_default", "zero")
+__all__ = ("commonmark", "default", "gfm_like", "gfm_like2", "js_default", "zero")
 
 from ..utils import PresetType
 from . import commonmark, default, zero
@@ -25,4 +25,23 @@ class gfm_like:  # noqa: N801
         config["components"]["inline"]["rules2"].append("strikethrough")
         config["options"]["linkify"] = True
         config["options"]["html"] = True
+        return config
+
+
+class gfm_like2:  # noqa: N801
+    """GitHub Flavoured Markdown (GFM) like, extended.
+
+    Builds on ``gfm-like`` and additionally enables:
+
+    - Task lists (``- [x] done``)
+    - Alerts (``> [!NOTE]``)
+    - Single-tilde strikethrough (``~text~`` in addition to ``~~text~~``)
+    """
+
+    @staticmethod
+    def make() -> PresetType:
+        config = gfm_like.make()
+        config["options"]["tasklists"] = True
+        config["options"]["alerts"] = True
+        config["options"]["strikethrough_single_tilde"] = True
         return config
--- a/markdown_it/presets/__init__.py
+++ b/markdown_it/presets/__init__.py
@@ -42,6 +42,7 @@ class gfm_like2:  # noqa: N801
     def make() -> PresetType:
         config = gfm_like.make()
         config["options"]["tasklists"] = True
+        config["options"]["tasklists_editable"] = False
         config["options"]["alerts"] = True
         config["options"]["strikethrough_single_tilde"] = True
         return config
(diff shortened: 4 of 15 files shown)


chrisjsewell:
## Overview

This PR adds a new `gfm-like2` preset that extends `gfm-like` with three GitHub Flavored Markdown features: task lists, alerts, and single-tilde strikethrough. CI is passing. The change is well-structured with 12 files changed (+664/−22) and thorough test fixtures.

## Strengths

- **Sound architectural rationale** — The PR description clearly explains why these features belong in core rather than `mdit-py-plugins`: checkbox/alert detection needs to hook into `list_block()` and `blockquote()` internals during parsing, not as post-processing.
- **Backward compatible** — The existing `gfm-like` preset is untouched; all new behavior is opt-in via `gfm-like2` or individual option flags.
- **Good test coverage** — 39 fixture cases covering happy paths, edge cases (invalid syntax, case insensitivity, nesting, mixed items), and negative cases.
- **Clean state management** in `list.py` — `bMarks` is saved/restored properly around the checkbox advancement.

## Issues & Suggestions

### 1. Checkbox HTML is not `disabled` (Medium)
The rendered `<input type="checkbox">` is interactive by default. GitHub renders these as `disabled` to prevent user interaction in static markdown. Consider adding `disabled=""` to the checkbox:
```python
result += (
    '<input class="task-list-item-checkbox"'
    f' type="checkbox" disabled=""{checked_attr}> '
)
```

### 2. Alert: `contentStart < nextLine` guard may skip `state.line` assignment (Medium)
In `blockquote.py`, when `contentStart >= nextLine` (empty alert body), the code sets `state.line = nextLine` in the `else` branch. But when `contentStart < nextLine`, `state.md.block.tokenize()` is called — verify that `tokenize()` always updates `state.line` to `nextLine`. If not, this could leave `state.line` in an inconsistent state. The existing blockquote code path relies on `tokenize` doing this, so it's likely fine, but worth a comment.

### 3. Alert detection only checks `startLine`, not the first *content* line (Low)
`_detect_alert` is called with `startLine`, which should be the first line of blockquote content (after `>` stripping). The guard `nextLine > startLine` ensures at least one line was parsed. This looks correct, but the variable naming (`startLine` being both the blockquote start and the content line after prefix stripping) is confusing — a comment clarifying this would help.

### 4. `_detect_task_checkbox` requires trailing whitespace, but `checkboxLen` handles missing space (Low)
The function requires `pos + 4 > maximum` to fail, meaning it needs at least `[ ] ` with a space. But `checkboxLen` conditionally adds the 4th char. Since `_detect_task_checkbox` already enforces the space exists, `checkboxLen` will always be 4 in practice. The conditional at lines checking `contentStart + 3` is dead code — consider simplifying to always set `checkboxLen = 4`.

### 5. Strikethrough: 3+ tildes silently consumed as text (Low)
When `single_tilde` is enabled and ≥3 tildes are encountered, they're pushed as a plain text token and `True` is returned. This means the strikethrough rule "claims" the tildes, preventing any other inline rule from processing them. This matches GitHub behavior but is worth documenting as intentional.

### 6. Missing `disabled` attribute in test fixtures
If you add `disabled=""` per suggestion #1, all tasklist fixture expected outputs would need updating.

### 7. Preset naming
`gfm-like2` is functional but not very descriptive. Consider `gfm-like-extended` or `gfm-like-full`? This is subjective — the current name works if you prefer brevity.

## Summary

This is a well-implemented, well-tested PR with a clear rationale. The main actionable item is adding `disabled` to checkbox inputs (#1). The rest are minor cleanups. The approach of integrating into existing block parsers rather than post-processing is the right call architecturally.

chrisjsewell:
## Changes Since Last Review

The PR now has 3 commits, 14 files changed (+680/−24). The new commits address the previous review feedback:

### ✅ Previously Raised Issues — Now Resolved

1. **Checkbox `disabled` attribute** — Fixed. Checkboxes now render with `disabled=""` by default, controlled by a new `tasklists_enabled` option that omits it when `True`. All test fixtures updated accordingly.

2. **`startLine` comment clarity in blockquote.py** — Fixed. A clear comment now explains that `startLine` refers to the first content line after `>` prefix stripping.

3. **`tokenize()` / `state.line` contract comment** — Fixed. An inline comment now documents that `tokenize()` updates `state.line` to `nextLine`.

4. **Dead code in `checkboxLen`** — Fixed. The conditional logic is gone; `checkboxLen` is now always set to `4`, with a comment explaining `_detect_task_checkbox` already guarantees trailing whitespace.

5. **3+ tildes documentation** — Fixed. The docstring and inline comment now explicitly note this is intentional GitHub-matching behavior.

### ✅ New: Documentation Updates (commit 3)

- **`docs/plugins.md`** — Lists task lists and alerts as embedded plugins; shows `gfm-like` vs `gfm-like2` usage side by side.
- **`docs/using.md`** — Documents the `gfm-like2` preset in the configuration section.
- **`utils.py`** — Added `tasklists_enabled` to `OptionsType` with docstring.

---

## Remaining Minor Items

### 1. `tasklists_enabled` naming (Nit)
The option name `tasklists_enabled` reads like "are task lists enabled" rather than "are checkboxes interactive." Something like `tasklists_interactive` or `tasklists_editable` would be clearer. This is very minor — the docstring does explain it.

### 2. `tasklists_enabled` not set in `gfm-like2` preset
The `gfm_like2` preset sets `tasklists=True` but doesn't set `tasklists_enabled`. This means the default is `False` (disabled/readonly checkboxes), which matches GitHub's static rendering. This is correct behavior, just noting it's implicit via `options.get("tasklists_enabled", False)` in the renderer rather than explicit in the preset.

### 3. No test for `tasklists_enabled=True` path
There's no fixture test verifying that setting `tasklists_enabled=True` produces checkboxes *without* `disabled=""`. A single test case would cover this code path.

### 4. Alert `lines` variable unused in alert branch
In blockquote.py, the alert branch does `token.map = lines = [startLine, 0]`, but `lines[1]` is never updated in the alert path (unlike the regular blockquote path where `lines[1] = nextLine`). This means `alert_open.map` will always be `[startLine, 0]` — likely a bug. The regular blockquote path updates `lines[1]` later. You should add `lines[1] = nextLine` before `alert_close`.

---

## Verdict

The review feedback was thoroughly addressed. The PR is in good shape. **Item #4 (alert `map` not updated)** is the only functional issue I'd flag as needing a fix before merge. The rest are minor polish items.

PR #389, opened 2026-05-04
👌 Fix quadratic complexity in `fragments_join` / `text_join`

petricevich:
Optimize adjacent-token joining in both inline cleanup stages by replacing repeated pairwise string concatenation with a single `"".join(...)` over each contiguous run.

## Details

- `fragments_join` merges adjacent `text` tokens left behind after emphasis/strikethrough post-processing and recalculates token levels
- `text_join` converts `text_special` tokens to `text` and performs the final adjacent-text merge in the inline token stream

Both rules previously rebuilt growing strings incrementally, which can become quadratic for long runs.

## Why

Tested on an adversarial ~190 KB document with ~30k intraword underscores on a single line. With `tracemalloc` running:

|           | render time | peak Python alloc |
|-----------|-------------|-------------------|
| before    | 2.2s        | 4476 MB           |
| after     | 0.6s        | 23 MB             |

It's not just a contrived attack input - this kind of thing also shows up naturally in Markdown produced by OCR pipelines, where tables of identifiers / references can easily contain very long runs of underscores or other delimiter characters.

## Tests

Added focused tests for both rules:

- `fragments_join`: verifies raw adjacent text fragments remain when both join stages are disabled, and that `fragments_join` alone collapses them when `text_join` is disabled
- `text_join`: verifies escaped characters remain as multiple `text_special` tokens when `text_join` is disabled, and are converted and merged into a single `text` token when enabled

## Result

No behavioral change in parser output, with less unnecessary work when joining long runs of adjacent tokens.

### DIFF
--- a/markdown_it/rules_core/text_join.py
+++ b/markdown_it/rules_core/text_join.py
@@ -21,7 +21,10 @@ def text_join(state: StateCore) -> None:
 
         # convert text_special to text and join all adjacent text nodes
         new_tokens: list[Token] = []
-        for child_token in inline_token.children or []:
+        children = inline_token.children or []
+        i = 0
+        while i < len(children):
+            child_token = children[i]
             if child_token.type == "text_special":
                 child_token.type = "text"
             if (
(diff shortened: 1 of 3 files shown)


chrisjsewell:
Thanks, will double check soon, but sounds good in principle

chrisjsewell:
I added an extra commit [4a89f1d](https://github.com/executablebooks/markdown-it-py/pull/389/commits/4a89f1da9ef2d64bab52156b2f8e85179f0ad9a2) and updated the PR title / description, hope that's all good 😅

chrisjsewell (approved):

PR #404, opened 2026-06-21
makes some changes in order to build docs quickly

gabrielfalcao:
the point of these changes is to enable a small set of commands to build the docs. only requirement is to have [uv](https://docs.astral.sh/uv/) pre-installed:

```shell
cd docs && make html
```

the is a bit of [JQ](https://jqlang.org/)-like kung-fu done via [jaq](https://crates.io/crates/jaq) which is a drop-in replacement for JQ that supports other formats beyond json: toml, yaml [and so on](https://github.com/01mf02/jaq/blob/56a57f0b4f7c5e06cc3f689b185889874d6f1fbf/jaq-fmts/Cargo.toml#L13-L21)

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -4,15 +4,16 @@
 # You can set these variables from the command line, and also
 # from the environment for the first two.
 SPHINXOPTS    ?=
-SPHINXBUILD   ?= sphinx-build
+SPHINXBUILD   ?= uv run sphinx-build
 SOURCEDIR     = .
-BUILDDIR      = _build
+BUILDDIR      = build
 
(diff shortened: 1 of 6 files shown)


chrisjsewell:
Closing this one: the Read the Docs build fails on this branch, `docs/walkthru.sh` is a raw shell-history dump rather than a script, `docs/pyproject.toml` sets `requires-python >=3.14` while the project supports 3.10+, and it pins myst-parser to an unmerged branch. If there is a concrete docs-build problem to fix, a focused PR with a description would be very welcome.

---

Issue #377, opened 2025-12-12
HTML comment block incorrectly ended before `-->` when rendering CommonMark


Jayman2000:
### Describe the bug

**context**
I’m trying use markdown-it-py to convert this chunk of Markdown (taken from [here](https://github.com/DescentDevelopers/Descent3/blob/0171d7b76a0c6d77908169af8ff3ea8d4d1cd930/BUILD.md#building---windows)) into HTML:

````markdown
1. **Install the prerequisite build tools.**

    Make sure that you have Git and Visual Studio 2022 with the “Desktop development with C++” workload and the “C++ MFC for latest v143 build tools (x86 & x64)” component. If you don’t already have those installed or you aren’t sure, then open an elevated Command Prompt and run:
    <!--
    The following code block specifies the full path to the Visual Studio Installer because the Visual Studio Installer doesn’t add itself to the user’s Path. The installer is guaranteed to be in a specific location on 64-bit systems [1]. The installer will be in a different location on 32-bit systems [2], but Visual Studio 2022 doesn’t support 32-bit systems [3] so we can ignore that detail.

    [1]: <https://learn.microsoft.com/en-us/visualstudio/install/use-command-line-parameters-to-install-visual-studio?view=vs-2022>
    [2]: <https://github.com/microsoft/vswhere/wiki#installing>
    [3]: <https://learn.microsoft.com/en-us/answers/questions/1689898/does-visual-studio-build-tools-2022-support-32-bit>
    -->
    ```batch
    winget install Git.Git Microsoft.VisualStudio.2022.Community

    "%ProgramFiles(x86)%\Microsoft Visual Studio\Installer\setup.exe" modify^
        --passive^
        --channelId VisualStudio.17.Release^
        --productId Microsoft.VisualStudio.Product.Community^
        --add Microsoft.VisualStudio.Workload.NativeDesktop;includeRecommended^
        --add Microsoft.VisualStudio.Component.VC.ATLMFC
    ```
````

**expectation**
I expect markdown-it-py to generate HTML that contains `-->`.

**bug**
markdown-it-py actually generates HTML that does not contain `-->`:

```html
<ol>
<li>
<p><strong>Install the prerequisite build tools.</strong></p>
[...]

PR #406, opened 2026-06-25
🐛 FIX: html_block: blank lines inside containers prematurely terminate non-blank-line-terminator sequences

binggao1230:
## Problem

HTML blocks started with `<!--`, `<?`, `<!A`, `<![CDATA[`, or `<script>`/`<pre>`/`<style>`/`<textarea>` (CommonMark types 1–5) have **specific** end conditions (`-->`, `?>`, `>`, `]]>`, closing tags). Per [CommonMark spec §4.6](https://spec.commonmark.org/0.31.2/#html-blocks), these blocks continue until the matching end tag is found, or the last line of the document/container — **blank lines inside them do not terminate them**.

When such a block appeared inside a container (e.g. a list item), a blank line within the comment caused early termination. The scan loop in `html_block` checks `state.sCount[nextLine] < state.blkIndent` to detect container boundaries, but blank lines always have `sCount = 0`, so any positive `blkIndent` (set by the list rule) incorrectly fired the guard.

**Reproduction** (from #377):

```python
md = MarkdownIt()
text = '''1. item

    <!--
    The following code block specifies the full path.

    [1]: <https://example.com>
    -->
    paragraph
'''
print(md.render(text))
```

Before this fix, `-->` was escaped as `--&gt;` inside a paragraph (comment prematurely closed at the blank line).

## Fix

One-character addition: skip the `sCount < blkIndent` guard when the line is blank (`state.isEmpty(nextLine)`). Blank lines in sequences 6–7 (end condition `^$`) still terminate correctly because an empty `lineText` matches `^$` through the existing end-condition check.

## Test

New fixture in `commonmark_extras.md` covering the exact case from #377.

This pull request was prepared with the assistance of AI, under my direction and review.

### DIFF
--- a/markdown_it/rules_block/html_block.py
+++ b/markdown_it/rules_block/html_block.py
@@ -68,7 +68,7 @@ def html_block(state: StateBlock, startLine: int, endLine: int, silent: bool) ->
     # Let's roll down till block end.
     if not html_seq[1].search(lineText):
         while nextLine < endLine:
-            if state.sCount[nextLine] < state.blkIndent:
+            if state.sCount[nextLine] < state.blkIndent and not state.isEmpty(nextLine):
                 break
 
             pos = state.bMarks[nextLine] + state.tShift[nextLine]
--- a/tests/test_port/fixtures/commonmark_extras.md
+++ b/tests/test_port/fixtures/commonmark_extras.md
@@ -749,3 +749,27 @@ Issue #204. Combination of blockquotes, list and newlines causes an IndexError
 </li>
 </ul>
 .
+
+Issue #377. HTML comment block inside list item should not end at blank line
+.
+1. item
+
+    <!--
+    comment with
+
+    blank line inside
+    -->
+    paragraph
+.
+<ol>
+<li>
+<p>item</p>
+ <!--
+ comment with
+
+ blank line inside
+ -->
+<p>paragraph</p>
+</li>
+</ol>
+.


chrisjsewell:
Thanks @binggao1230, and apologies for the slow response. I agree the fix is right against the CommonMark spec: a blank line inside a list item does not end the container, so a type 1–5 HTML block should continue to its real end condition, and the reference implementation renders the #377 example as a comment.

The blocker is that this port deliberately mirrors the JavaScript markdown-it, and markdown-it 14.1.1 produces byte-identical output to markdown-it-py for this input (with `html: true`). The same guard, `state.sCount[nextLine] < state.blkIndent`, is in upstream `html_block.mjs`, so this is an upstream bug rather than a porting error, and it was raised there as markdown-it/markdown-it#1144. Diverging here would mean the two implementations disagree on HTML block termination, which tends to come back to us as bug reports from projects that rely on parity.

So I would rather not merge this ahead of upstream. The change is a one-liner, and your fixture translates directly, so would you be willing to open the equivalent pull request against markdown-it/markdown-it, referencing #1144? As soon as it lands there we will port it here, crediting you. I will leave this PR open in the meantime so it is not lost.

binggao1230:
Thanks for the careful parity check. It turns out markdown-it/markdown-it#1155 already fixed #1144 and was merged as 68cfb8c on May 22; current markdown-it master contains its endsOnBlankLine guard. Version 14.1.1 predates that merge, which explains the byte-identical release behavior. I'll rebase this PR and align it with the upstream implementation and its list/blockquote fixtures.

PR #411, opened 2026-07-29
👌 Fix quadratic inline tokenization on runs of special characters

hdimer:
## Summary

Long runs of characters that begin an inline rule but do not form a construct — bare `&` / `<`, incomplete entities like `&am&am…`, etc. — were tokenised in **O(n²)** time. A single such character is fine; it is the *count* that blows up. For example, `md.render("&" * 128_000)` took over a second and grew quadratically.

This is the same class as the previously fixed #367 and #389.

## Root causes

Two independent quadratic factors, both hit once per character in these runs:

1. **`state.pending` accumulation.** The inline tokenizer's fallback path does `state.pending += state.src[state.pos]` one character at a time. Because `pending` is an attribute, `str += ch` cannot use CPython's in-place concatenation optimisation (the attribute keeps a second reference), so each append rebuilds the whole string — O(len) per character.
2. **`src[pos:]` slicing in `entity` / `html_inline`.** These rules matched `^`-anchored regexes against `state.src[pos:]`, copying the remaining source on every `&` / `<`.

In JavaScript `markdown-it` neither is quadratic (rope-backed string concat, and the equivalent regexes are applied with a sticky/lastIndex match), so this is Python-port-specific.

## Fix

- `StateInline.pending` now accumulates through a lazily-materialised list buffer (`append_pending`), giving amortised O(1) appends. It still presents as a plain `str` to every reader (the getter joins the buffer on demand and caches).
- `entity` and `html_inline` anchor their regexes with `.match(state.src, pos)` instead of slicing (the leading `^` is dropped, since `.match` already anchors at `pos`). `HTML_TAG_RE` is only used by `html_inline`, so this is contained.

Rendering is now linear in all these cases (measured slope ≈ 1.0, including the `html=True` `html_inline` path); `"&" * 500_000` drops from seconds to well under one. **Output is unchanged** — the full test suite passes.

## Tests

Added `test_long_special_char_runs_stay_linear` in `tests/test_api/test_main.py`: it asserts correct rendering for the affected constructs and renders a large input whose previous O(n²) behaviour would exceed the global 10s test timeout. Full `pytest tests/` passes (982). `ruff check` / `ruff format` (0.15.12) and `mypy` (1.20.2, strict) are clean.

---

Disclosure: this fix was developed with AI assistance; I reviewed it and stand behind it.

### DIFF
--- a/markdown_it/common/html_re.py
+++ b/markdown_it/common/html_re.py
@@ -21,7 +21,9 @@ declaration = "<![A-Za-z][^>]*>"
 cdata = "<!\\[CDATA\\[[\\s\\S]*?\\]\\]>"
 
 HTML_TAG_RE = re.compile(
-    "^(?:"
+    # No leading `^`: this is applied with `.match(src, pos)`, which anchors at
+    # `pos` without slicing `src[pos:]` (O(len) per call, quadratic over a run).
+    "(?:"
     + open_tag
     + "|"
     + close_tag
--- a/markdown_it/parser_inline.py
+++ b/markdown_it/parser_inline.py
@@ -197,7 +197,7 @@ class ParserInline:
                     break
                 continue
 
-            state.pending += state.src[state.pos]
+            state.append_pending(state.src[state.pos])
             state.pos += 1
 
         if state.pending:
--- a/markdown_it/rules_inline/backticks.py
+++ b/markdown_it/rules_inline/backticks.py
@@ -25,7 +25,7 @@ def backtick(state: StateInline, silent: bool) -> bool:
 
     if state.backticksScanned and state.backticks.get(openerLength, 0) <= start:
         if not silent:
-            state.pending += marker
+            state.append_pending(marker)
         state.pos += openerLength
         return True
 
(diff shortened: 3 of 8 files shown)


chrisjsewell (approved):
thanks!

chrisjsewell:
Thanks @hdimer, and apologies this sat for a while. You identified both root causes and the right fix here, and #423 (now merged) carries your change verbatim as its first commit, with you credited as co-author.

While auditing it before merging we found one gap that #423 adds on top: the buffered `pending` getter materialised with `self._pending += joined`, which is itself an attribute concatenation, so any inline rule that reads `state.pending` on every character (the mdit-py-plugins `attrs` rule does, and myst-parser enables it) re-introduced the quadratic. The getter now moves the string into a local before concatenating so CPython can resize it in place. #423 also hardens the setter and `copy.copy`, and replaces the timeout-based tests with deterministic ones, since CI runs everything under coverage.

Closing this in favour of #423. It will be in the next release. Much appreciated!

PR #417, opened 2026-08-13
🐛 FIX: keep the hard line break after a literal backslash

eeshsaxena:
markdown-it-py renders a literal backslash immediately before a two-space hard line break as a soft break with a stray space, rather than a hard break:

```python

MarkdownIt("commonmark").render("foo\\  \nbar")
# markdown-it-py: '<p>foo\\ \nbar</p>\n'
# markdown-it:    '<p>foo\\<br />\nbar</p>\n'
```

markdown-it's `escape` rule has a dedicated case for this: when a `\` is followed by a space it emits just the backslash and leaves the space unconsumed, so the trailing two-space hard break is still detected by the newline rule (its `escape.mjs` even carries a comment to that effect). This port jumped straight to the generic escape branch, which consumed the space and left only one space before the newline, so the hard break was downgraded to a soft break.

This ports the missing space case so the output matches markdown-it. I added a regression test in `test_misc.py`, and the `test_cmark_spec` and `test_port` suites still pass.

### DIFF
--- a/markdown_it/rules_inline/escape.py
+++ b/markdown_it/rules_inline/escape.py
@@ -36,6 +36,17 @@ def escape(state: StateInline, silent: bool) -> bool:
         state.pos = pos
         return True
 
+    # '\' before a space is a literal backslash. Don't consume the space, so a
+    # trailing two-space hard line break is still detected by the newline rule.
+    if ch1 == " ":
+        if not silent:
+            token = state.push("text_special", "", 0)
+            token.content = "\\"
+            token.markup = "\\"
+            token.info = "escape"
+        state.pos = pos
+        return True
+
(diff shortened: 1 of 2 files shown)


eeshsaxena:
Hi! Gentle nudge on this one whenever you have some bandwidth. It's a small, self-contained fix (`🐛 FIX: keep the hard line break after a literal backslash`), and it's currently mergeable with no conflicts. No urgency at all, and I'm happy to make any changes you'd like. Thanks for maintaining markdown-it-py!

eeshsaxena:
Closing this to clear out my older open PRs. Nothing wrong with the change on my side, I am just tidying up a backlog. If it is still useful to you, say so and I will reopen it and rebase it on current main.

PR #421, opened 2026-09-06
📚 DOCS: Document renderer constructor differences

### LINKED ISSUE #127: Document renderer class init difference in `port.yaml`
In JS, the renderer class is initialized with no arguments. In `markdown-it-py` renderer classes take a `MarkdownIt` parser instance as argument.

user01010111:
Document that `MarkdownIt` passes itself to `renderer_cls`, unlike the argument-free JavaScript renderer constructor. Clarify the custom constructor contract and that `RendererHTML` accepts but ignores an optional parser argument.

Closes #127.

Validation: 981 tests passed; all pre-commit checks passed; strict docs build passed on Python 3.10 with Sphinx 8.1.3. Sphinx 9 fails in the unchanged API-doc configuration on both upstream and this branch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change Log
 
+## Unreleased
+
+* 📚 Document the Python renderer constructor contract.
+
 ## 4.2.0 - 2026-05-07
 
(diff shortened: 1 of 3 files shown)


chrisjsewell (approved):
thanks

Issue #172, opened 2021-09-06
Hard line breaks parsed incorrectly in version 1.1.0


jamesquilty:
### Describe the problem

I found what I think is a bug with the parsing of hard like breaks <https://spec.commonmark.org/0.30/#hard-line-break>. [Example 634](https://spec.commonmark.org/0.30/#example-634)

```markdown
foo\
baz
```

should be parsed as

```html
<p>foo<br />
baz</p>
``` 

but I receive

```html
<p>foo\</p>
<p>baz</p>
```

The markdown-it demo <https://markdown-it.github.io/> produces output consistent with [Example 634](https://spec.commonmark.org/0.30/#example-634), so I think this is a bug in markdown-it-py only.

### Steps to reproduce

```shell
$ markdown-it
markdown-it-py [version 1.1.0] (interactive)
Type Ctrl-D to complete input, or Ctrl-C to exit.
>>> foo\
... baz
... ^D
<p>foo\</p>
<p>baz</p>
```

### The version of Python you're using

Python 3.7.3

### Your operating system

macOS

chrisjsewell:
Heya, thanks. I just tested this locally, and looks to be only an issue with the interactive CLI,
i.e. if I write:

`test.txt`:
```
foo\
baz
```

you get

```console
$ markdown-it test.txt
<p>foo<br />
baz</p>
```

but then with the interactive "mode" I recreate your findings

jamesquilty:
Thanks for taking the time to reproduce this. It's odd that it only affects the interactive interface, but it's good that it's less serious than if it affected the core code. I'll have to be careful about testing with the CLI in the future!

chrisjsewell:
Yeh must just be a weird quirk of the CLI stdin, but I'll see if I can work it out at some point

jamesquilty:
I just tried reproducing the test case with a file and received different results to those you show. I receive:

```shell
$ markdown-it test.md
<p>foo\
baz</p>
```

i.e. the `<br />` which should be present at the end of the first line has been rendered as a newline character.

chrisjsewell:
to check, what version of markdown-it-py and also perhaps check the encoding of the file (although I'm on a mac as well, so should not be a problem)

jamesquilty:
[...]

PR #422, opened 2026-09-06
✨ NEW: Add --enable-tables to the CLI

### LINKED ISSUE #352: [3.0.0] Command `markdown-it` does not allow activating support for tables? :cry:
Hi! 👋 

I just learned that `markdown-it README.md > README.htm` does not render tables, that `markdown-it --help` does not show any related switches and that I need something like…

```bash
python3 -c 'import sys; from markdown_it import MarkdownIt; md = MarkdownIt("commonmark").enable("table"); print(md.render(sys.stdin.read()))' < README.md > README.htm
```

…to render with support for tables active. Is there a chance that maybe a commend line argument `--enable-tables` could be added?

Thanks and best, Sebastian

user01010111:
Closes #352.

Add `--enable-tables` for files, standard input and interactive input, using `MarkdownIt.enable("table")`. Default CommonMark rules and HTML settings remain unchanged.

Pass the configured parser through each CLI route and preserve interactive input line breaks so tables can be entered line by line. Update the CLI help, README and changelog.

Validation: 987 tests passed on Python 3.13, including default/opt-in checks across all CLI routes; all pre-commit hooks passed. Source distribution and wheel builds passed, with file/stdin checks against the installed command. Strict docs passed on Python 3.10 with Sphinx 8.1.3. Newer Sphinx versions hit the existing `apidoc.OPTIONS` incompatibility.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change Log
 
+## Unreleased
+
+* ✨ Add `--enable-tables` to the CLI for file, standard input and interactive parsing.
+
 ## 4.2.0 - 2026-05-07
 
 * ✨ Add `make_fence_rule()` factory for configurable fence markers in [#394](https://github.com/executablebooks/markdown-it-py/pull/394)
--- a/README.md
+++ b/README.md
@@ -101,17 +101,18 @@ Render markdown to HTML with markdown-it-py from the
 command-line:
 
 ```console
-usage: markdown-it [-h] [-v] [--stdin|filenames [filenames ...]]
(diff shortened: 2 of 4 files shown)


chrisjsewell:
Thanks @user01010111! Your commit has been merged verbatim, with your authorship preserved, via #425 (we could not push to this fork branch to bring it up to date). #425 also added a changelog line and a regression test for the interactive-mode line-joining fix that was part of this change, which turned out to be the root cause of the five-year-old #172. Closing this in favour of #425; it will be in the next release.

